Issue #36, opened 2025-04-07
Make Simulator Generic For Different Attacks


carlaKC:
Right now, the sink attack is quite intertwined with the simulator. It should be more generic so that it can be used to try out different attacks.

The simulator needs to be run with the following parameters:
- Alias of target node
- Aliases of the attacking nodes
- Attack uuid

Right now we require that the simulation is run an attacker bootstrap file to build up the attacker's reputation on its channels. Some attacks won't require ths "passive" reputation bootstrap where the attacker just sits in the network and forwards whatever they get, so we should make this behavior optional so that the attacker can build their reputaiton however they please. 

Then each attack will be required to define:
- A uuid for the attack
- A list of channels that should be general jammed for the attack
- A closure that runs the attack (we need access to underlying `sim-node`s for this)
- A closure that implements interception for the attacking nodes
- A closure that provides shutdown conditions for the attack (the revenue check will be standard)

carlaKC:
Addressed in #41

PR #37, opened 2025-04-09
prefactor: prepare for pulling attack out into module

carlaKC:
In preparation for #36, do some renaming and refactoring.
* We call things `incoming_revenue` in a lot of places, which is confusing when we're actually tracking this bidirectinally.
* Rename sink-specific files/structs
* Move to a more generic `list_channels` vs `list_reputation_pairs`

### DIFF
--- a/ln-simln-jamming/src/attack_interceptor.rs
+++ b/ln-simln-jamming/src/attack_interceptor.rs
@@ -6,8 +6,7 @@ use bitcoin::secp256k1::PublicKey;
 use ln_resource_mgr::{EndorsementSignal, ForwardingOutcome, HtlcRef, ProposedForward};
 use simln_lib::clock::Clock;
 use simln_lib::sim_node::{ForwardingError, InterceptRequest, InterceptResolution, Interceptor};
-use simln_lib::ShortChannelID;
-use std::collections::HashMap;
+use std::collections::HashSet;
 use std::error::Error;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


carlaKC:
@elnosh the last commit here is a WIP version of removing `list_channel_pairs` with a simpler building block, moving the pairs logic into main. Might clean this up a bit more, but wanted to get it up so to give an idea of what the API would look like.

Also done a few renames of badly named "incoming" - no wonder that was confusing!

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -158,7 +158,7 @@ impl ForwardManagerImpl {
                 forward.incoming_ref.channel_id,
             ))?;
 
-        let incoming_threshold = incoming_channel
+        let revenue_threshold = incoming_channel
```
elnosh, line 161: The rename from `incoming` to bidirectional in other places made sense but in this method where we are getting the forwarding outcome I thought it'd be good to maintain the `incoming` because when we implement incoming reputation we will need the revenue on the outgoing channel so we'd need to differentiate those two. Or do you think that would be somewhere else?

elnosh (commented):
the new `list_channels` in the `ReputationMonitor` makes sense. 

Although not sure about the network reputation count in main. I wasn't able to build the attacker reputation. I ran it with the same network and `bootstrap` file but got `Error: "attacker has 0/90 good reputation pairs which does not meet threshold"`

ln-resource-mgr/src/forward_manager.rs
elnosh: ah nice, I saw it having two `ReputationCheck`s but this looks better.

carlaKC:
I've merged some of the trivial refactors and updated the approach here to better set us up for making the simulator more generic.

This should be relatively stable now @elnosh, but sorry for the refactor headaches this will give you!

Issue #31, opened 2025-03-28
Startup is very slow


carlaKC:
When we start up the simulation, we read two large files that contain channel forwards in them:
* `bootstrap`: contains forwards for the full network, used to generate reputation histories for each node
* `peacetime`: contains forwards for the full network, used to project what our peacetime revenue would be

These are read and then filtered in a very lazy way, so the simulation can startup quite slow (eg, files can be around 1GB for 1 year of data for 50 node networks).

There's definitely something smarter we can do here, but I haven't looked into specific solutions.

carlaKC:
Adding a bit more context to why this is a problem.

## Bootstrap

One of the most commonly changed parameters when we're running simulations is `--attacker-bootstrap` because we want to see how different attacks play out with different levels of attacker reputation (longer bootstrap / greater reputation). For this reason, we need to re-process the bootstrap file to build up different levels of attacker reputations.

That said, we do have quite fixed values that we'll run this for - eg, 15 days / 40 days / 90 days for the same file. So there's some possibility that we could cache the end state of reputation for the network + attacker and save it to disk (creating a `/cache` dir that stores reputation data more succinctly) and opportunistically check whether it's there on startup. We'd need to make sure that we're _definitely_ loading the right cache (because we pretty much always have the same name for the `bootstrap.csv` but the contents may have changed). This is probably achievable by checking the last modified timestamp of the file and saving that in the cached file's name.

## Peacetime

This is a pretty straightforward optimization - right now we read the whole file and replay it in memory, we could just read chunks of it at a time as we need it.

carlaKC:
Chatted about this offline, decided to move filtering of reputation into a pre-processing step that provides summaries of reputation scores and then import them so that people don't need to reprocess every time.

Rather than providing a set of HTLC forwards for the simulator to read in, we'll generate the reputation values for the network at a point in time and pass that to the simulator.

## Step 1: CLI tool option to build reputation

Parameters:
- Path to file with htlc forwards
- Optionally provide attacker alias + bootstrap period to filter out a node for a period of time
- Path to output file (have a smart default if this isn't set)

We want to be able to generate these files without the attacker filtering as well, because some attacks will just want to start with an honest network of peers with established reputation.

This tool should then run the same code as in `get_history_for_bootstrap` and `bootstrap_network_history` to build up reputation state in an internal `ForwardManager`. To write this to disk, we can use `list_reputation_pairs` to get the reputation scores for each node's channels.

This should then be written to disk as:
[...]

PR #38, opened 2025-04-10
reputation builder cli

elnosh:
partially work towards https://github.com/carlaKC/jam-ln/issues/31. It depends on https://github.com/carlaKC/jam-ln/pull/37 for the new `list_channels` method on the `ReputationMonitor`. 

Changes done here:
- break up the reading of both bootstrap and peacetime files into chunks and read them in different tasks.
- add cli tool that will read forwards from a bootstrap file to build the reputation of a network.
     - writes a snapshot of the reputation state of the network to disk to then use as starting point for the simulation.


elnosh:
rebased and pulled new changes from https://github.com/carlaKC/jam-ln/pull/37

ln-simln-jamming/src/parsing.rs
```diff
@@ -173,96 +176,196 @@ fn parse_duration(s: &str) -> Result<Duration, String> {
...
-        };
+            Ok(forwards)
+        }));
+    }
 
-        forwards.push(forward);
+    let mut forwards = Vec::new();
+    for task in tasks {
+        let task_forwards = task.await??;
+        forwards.extend_from_slice(&task_forwards);
```
carlaKC, line 331: We're no longer returning forwards in the order that they're written in the file, worth making a note of this in the function's doc (seems like the sort of thing you'd be surprised _doesn't_ happen)

Would also be nice to have a meta-note in the commit that we have to sort by add/remove ts _anyway_ later on so it's not worth the expense of sorting here.

ln-simln-jamming/src/parsing.rs
```diff
@@ -173,96 +176,196 @@ fn parse_duration(s: &str) -> Result<Duration, String> {
...
-        heap.push(RevenueEvent {
-            timestamp_ns: record[3].parse()?,
-            fee_msat: incoming_amt - outgoing_amt,
-        })
+    while let Some(res) = tasks.join_next().await {
+        res??
     }
 
+    // TODO: is this bad?
+    let heap = heap.lock().unwrap().clone();
```
carlaKC, line 392: Cloning the underlying heap is going to be really expensive for larger files.

`Arc::try_unrwap` + `Mutex::into_inner` should give you access to the underlying heap without copying.

ln-simln-jamming/src/parsing.rs
```diff
@@ -173,96 +176,196 @@ fn parse_duration(s: &str) -> Result<Duration, String> {
         .map_err(|e| format!("Invalid duration '{}': {}", s, e))
 }
 
+fn find_next_newline(file: &mut File, start: u64) -> Result<u64, BoxError> {
+    let mut position = start;
+    file.seek(std::io::SeekFrom::Start(position))?;
+    let mut buffer = [0; 1];
```
carlaKC, line 206: Tempted to suggest a `BufReader` here to optimize these single byte reads a bit.

Perhaps not an issue because we have ~200 byte lines, so doing a few hundred single byte reads isn't so bad? Let's add a comment about the approach so that the single byte read doesn't look so scary to the next person who reads this :)

Cargo.toml
```diff
@@ -4,4 +4,5 @@ resolver = "2"
 members = [
     "ln-resource-mgr",
     "ln-simln-jamming",
+    "reputation-builder",
```
carlaKC, line 7: We're allowed to have multiple binary crates, how about this structure instead?

```
ln-simln-jamming/
├── Cargo.toml
└── src/
    ├── main.rs
    └── bin/
        ├── build_reputation.rs
```

Would allow us to de-dup some of the parsing defaults

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -15,6 +15,7 @@ use std::time::{Duration, Instant};
 /// Tracks reputation and revenue for a channel.
 #[derive(Debug)]
 struct TrackedChannel {
+    capacity_msat: u64,
```
carlaKC, line 18: Do we need to add `capacity_msat` here?

When we read in the reputation snapshot, we should have a graph alongside it which gives us channel policy information / capacity anyway (and if a channel is in the snapshot but not the graph then we should fail).

I'll also look at the follow up PR, maybe I'm missing something!

reputation-builder/src/main.rs
```diff
@@ -0,0 +1,161 @@
...
+                channel.0,
+                channel.1.capacity_msat,
+                channel.1.outgoing_reputation,
+                channel.1.bidirectional_revenue,
+            ))?;
+        }
+    }
+    csv_writer.flush()?;
+
+    log::info!("finished writing reputation snapshot");
```
carlaKC, line 158: nit: capitalize "Finished", perhaps also print the path we've written to.

carlaKC (commented):
Looking forward to this one! 🏎️

ln-simln-jamming/src/parsing.rs
elnosh: I feel very dumb

ln-simln-jamming/src/parsing.rs
elnosh: yeah thought it would be ok since it was for reads within a single line. I can change it tho

ln-resource-mgr/src/forward_manager.rs
elnosh: > Do we need to add `capacity_msat` here?

I added to then use it when calling `add_channel` on the `ReputationManager` since that methods takes in the channel capacity.  



> When we read in the reputation snapshot, we should have a graph alongside it which gives us channel policy information / capacity anyway

yeah, thought about that. Since I'm using the capacity like this when writing to the snapshot file: 

```
for pubkey in node_pubkeys {
        let channels = reputation_interceptor
            .list_channels(pubkey, InstantClock::now(&*clock))
            .await?;

        for channel in channels {
            csv_writer.serialize((
                pubkey,
                channel.0,
                channel.1.capacity_msat,
                channel.1.outgoing_reputation,
                channel.1.bidirectional_revenue,
            ))?;
        }
    }
```

I thought it would be easier to have the capacity on the `ChannelSnapshot` that is already being used and avoid doing the extra lookup to get the capacity from the graph

elnosh:
addressed some of the comments and now that incoming reputation has been merged, writing it to the snapshot file as well. Also, based on this comment: https://github.com/carlaKC/jam-ln/pull/42#discussion_r2052545135 now writing the undecayed revenue that we'll need to a file.

ln-simln-jamming/src/parsing.rs
```diff
@@ -181,96 +184,199 @@ fn parse_duration(s: &str) -> Result<Duration, String> {
...
+    let mut tasks: Vec<tokio::task::JoinHandle<Result<Vec<BootstrapForward>, BoxError>>> =
+        Vec::with_capacity(num_chunks);
+
+    for i in 0..num_chunks {
+        let start = breakpoints[i];
+        let end = if i == num_chunks - 1 {
+            file_size
+        } else {
+            breakpoints[i + 1]
+        };
```
carlaKC, line 282: Related to the above comment about how we create breakpoints, could this be simplified to just be `0...breakpoints.len() -1`? So that if we need fewer chunks than we thought, we just dispatch the number that we want?

ln-simln-jamming/src/parsing.rs
```diff
@@ -181,96 +184,199 @@ fn parse_duration(s: &str) -> Result<Duration, String> {
...
+    // Set breakpoints where the closest line ends.
+    let mut current_breakpoint = 0;
+    for _ in 0..num_chunks {
+        let mut end = current_breakpoint + chunk_size;
+        if end > file_size {
+            end = file_size;
+        }
+
+        breakpoints.push(current_breakpoint);
+        let next_eol_point = find_next_newline(&mut reader, end)?;
```
carlaKC, line 242: I think that this behaves a little unexpectedly if we hit the end of the file before we've used up all of our chunks? 

Say we have:
* `num_chunks` = 10
* `file_size` = 200
* Newlines at 101 and 182

We'll fill up our first few breakpoints as expected `[0, 102, 183]`, but once we hit the end of the file calling `find_next_newline(200)` will return `200` (nothing left to read) and then we fill the rest of the breakpoints with this value ending up with `[0, 102, 183, 201, 201, 201, 201, 201, 201, 201]`?

I think we can address this by adding a break if `current_breakpoint >= file_size` after we've `pushed` it?

Tried to [mock this out in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6f0ddbc720ab4141aa23903c50f94e36) to test this out without files.

This is probably more theoretical than practical, because I believe we only hit it when we _don't_ have a newline within our chunk (effectively letting us "skip" over one chunk, and ending up with an extra no-op one at the end).

ln-simln-jamming/src/bin/reputation_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let snapshot_dir = match cli.attacker_bootstrap {
+        // if a bootstrap period is provided, write results in ./reputation-snapshot/{duration}
+        Some(bootstrap_filter) => cli.results_dir.join(bootstrap_filter.0),
+        None => cli.results_dir,
+    };
+    fs::create_dir_all(&snapshot_dir)?;
+
+    let mut target_revenue = File::create(snapshot_dir.join("target-revenue.txt"))?;
+    write!(target_revenue, "{}", bootstrap_revenue)?;
```
carlaKC, line 209: Nice 👌

ln-resource-mgr/src/forward_manager.rs
carlaKC: Looking at `new_from_snapshot` in the follow up PR - the benefit that we get from writing the `capacity_msat` here is that we don't need to look the capacity up in `edges: &[NetworkParser]` - right?

I do think that we _might_ want to be cross-referencing the graph with the snapshot anyway (to prevent accidentally combining the wrong snapshot/graph) but happy to go ahead with this in the meantime. Get fancier when we need it..

carlaKC (commented):
One pedantic comment from me about having duplicate values in `breakpoints` if we get to the end of a file

Managed to hit this with a really small file + 6 workers locally: 
`breakpoints = [0, 178, 384, 590, 591, 591]`
```
incoming_amt,incoming_expiry,incoming_add_ts,incoming_remove_ts,outgoing_amt,outgoing_expiry,outgoing_add_ts,outgoing_remove_ts,forwarding_node,forwarding_alias,chan_in,chan_out
7947325,408,1739020833768692162,1739020833779307798,7946008,264,1739020833771365861,1739020833778509998,021998f701de16564da7abff57abf8a79bd722de3f0152d5e7a219021fd5287510,54,544258255814656,451899279081472
7946008,264,1739020833771365861,1739020833778509998,7944691,120,1739020833773044539,1739020833777719205,038a43948bc67703c9ce8a94aa298867b8533633ae95dc80edf6e34020e820f030,17,451899279081472,573945069764608
```

Not a trainsmash - we just spin up a few tasks that we don't really need, so not strictly blocking.

ln-resource-mgr/src/forward_manager.rs
elnosh: > Looking at `new_from_snapshot` in the follow up PR - the benefit that we get from writing the `capacity_msat` here is that we don't need to look the capacity up in `edges: &[NetworkParser]` - right?

yeah

ln-simln-jamming/src/parsing.rs
elnosh: yep, added a `break` if we are at the end of the file

carlaKC (approved):
Thanks for the final fixups!

PR #40, opened 2025-04-15
implement incoming reputation

elnosh:
This PR adds incoming reputation. 

When assessing an htlc, The `ForwardManager` now returns an `AllocationCheck` that includes values to check both incoming and outgoing reputation. 

The option to do either incoming or outgoing or both can be passed to the reputation and forward manager and it will do the checks accordingly when deciding to add a new htlc.

### DIFF
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -109,8 +109,41 @@ impl RevenueAverage {
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Reputation {
+    Incoming,
+    Outgoing,
+    Bidirectional,
+}
+
+impl Reputation {
+    pub fn sufficient_reputation(&self, check: &AllocationCheck) -> bool {
+        match self {
+            Reputation::Incoming => check
+                .reputation_check
+                .incoming_reputation
+                .sufficient_reputation(),
+            Reputation::Outgoing => check
+                .reputation_check
+                .outgoing_reputation
+                .sufficient_reputation(),
+            Reputation::Bidirectional => {
+                check
+                    .reputation_check
+                    .incoming_reputation
+                    .sufficient_reputation()
+                    && check
+                        .reputation_check
+                        .outgoing_reputation
+                        .sufficient_reputation()
+            }
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct ForwardManagerParams {
     pub reputation_params: ReputationParams,
+    pub reputation_check: Reputation,
     pub general_slot_portion: u8,
     pub general_liquidity_portion: u8,
     pub congestion_slot_portion: u8,
@@ -365,9 +398,11 @@ impl ReputationManager for ForwardManager {
 
         let allocation_check = inner_lock.get_forwarding_outcome(forward)?;
 
-        if let Ok(bucket) = allocation_check
-            .inner_forwarding_outcome(forward.amount_out_msat, forward.incoming_endorsed)
-        {
+        if let Ok(bucket) = allocation_check.inner_forwarding_outcome(
+            forward.amount_out_msat,
+            forward.incoming_endorsed,
+            self.params.reputation_check,
+        ) {
             inner_lock.htlcs.add_htlc(
                 forward.incoming_ref,
                 InFlightHtlc {
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -363,7 +363,7 @@ impl ReputationManager for ForwardManager {
             .lock()
             .map_err(|e| ReputationError::ErrUnrecoverable(e.to_string()))?;
 
-        // Remove the hltc from our tracker, as well as the incoming and outgoing direction's current state.
+        // Remove the htlc from our tracker, as well as the incoming and outgoing direction's current state.
         let in_flight = inner_lock
             .htlcs
             .remove_htlc(outgoing_channel, incoming_ref)?;
@@ -375,7 +375,7 @@ impl ReputationManager for ForwardManager {
                 incoming_ref.channel_id,
             ))?
             .incoming_direction
-            .remove_incoming_htlc(&in_flight, resolved_instant);
+            .remove_incoming_htlc(&in_flight, resolution, resolved_instant)?;
 
         inner_lock
             .channels
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -4,8 +4,8 @@ use crate::incoming_channel::IncomingChannel;
 use crate::outgoing_channel::{BucketParameters, OutgoingChannel};
 use crate::{
     AllocationCheck, BucketResources, ChannelSnapshot, ForwardResolution, HtlcRef, ProposedForward,
-    ReputationCheck, ReputationError, ReputationManager, ReputationParams, ResourceBucketType,
-    ResourceCheck,
+    ReputationCheck, ReputationError, ReputationManager, ReputationParams, ReputationValues,
+    ResourceBucketType, ResourceCheck,
 };
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
@@ -130,7 +130,7 @@ pub trait SimualtionDebugManager {
     fn general_jam_channel(&self, channel: u64) -> Result<(), ReputationError>;
 }
 
-/// Implements outgoing reputation algorithm and resource bucketing for an individual node.
+/// Implements incoming and outgoing reputation algorithm and resource bucketing for an individual node.
 #[derive(Debug)]
 pub struct ForwardManager {
     params: ForwardManagerParams,
@@ -150,7 +150,6 @@ impl ForwardManagerImpl {
     ) -> Result<AllocationCheck, ReputationError> {
         forward.validate()?;
 
-        // Get the incoming revenue threshold that the outgoing channel must meet.
         let incoming_channel = self
             .channels
             .get_mut(&forward.incoming_ref.channel_id)
@@ -158,36 +157,59 @@ impl ForwardManagerImpl {
                 forward.incoming_ref.channel_id,
             ))?;
 
-        let revenue_threshold = incoming_channel
+        let incoming_revenue_threshold = incoming_channel
             .bidirectional_revenue
             .value_at_instant(forward.added_at)?;
+
+        // Reputation from the incoming channel.
+        let incoming_reputation = incoming_channel
+            .incoming_direction
+            .incoming_reputation(forward.added_at)?;
 
         let no_congestion_misuse = incoming_channel
             .incoming_direction
             .no_congestion_misuse(forward.added_at);
 
         // Check reputation and resources available for the forward.
-        let outgoing_channel = &mut self
-            .channels
-            .get_mut(&forward.outgoing_channel_id)
-            .ok_or(ReputationError::ErrOutgoingNotFound(
-                forward.outgoing_channel_id,
-            ))?
-            .outgoing_direction;
+        let outgoing_channel = self.channels.get_mut(&forward.outgoing_channel_id).ok_or(
+            ReputationError::ErrOutgoingNotFound(forward.outgoing_channel_id),
+        )?;
+
+        // Revenue over the outgoing channel that the incoming channel must meet.
+        let outgoing_revenue_threshold = outgoing_channel
+            .bidirectional_revenue
+            .value_at_instant(forward.added_at)?;
+
+        // Reputation from the outgoing channel
+        let outgoing_reputation = outgoing_channel
+            .outgoing_direction
+            .outgoing_reputation(forward.added_at)?;
 
         Ok(AllocationCheck {
             reputation_check: ReputationCheck {
-                outgoing_reputation: outgoing_channel.outgoing_reputation(forward.added_at)?,
-                revenue_threshold,
-                in_flight_total_risk: self.htlcs.channel_in_flight_risk(
-                    ChannelFilter::OutgoingChannel(forward.outgoing_channel_id),
-                ),
-                // The underlying simulation is block height agnostic, and starts its routes with a height of zero, so
-                // we can just use the incoming expiry to reflect "maximum time htlc can be held on channel", because
-                // we're calculating expiry_in_height - 0.
-                htlc_risk: self
-                    .htlcs
-                    .htlc_risk(forward.fee_msat(), forward.expiry_in_height),
+                incoming_reputation: ReputationValues {
+                    reputation: incoming_reputation,
+                    revenue_threshold: outgoing_revenue_threshold,
+                    in_flight_total_risk: self.htlcs.channel_in_flight_risk(
+                        ChannelFilter::IncomingChannel(forward.incoming_ref.channel_id),
+                    ),
+                    // The underlying simulation is block height agnostic, and starts its routes with a height of zero, so
+                    // we can just use the expiry height to reflect "maximum time htlc can be held on channel", because
+                    // we're calculating expiry_in/out_height - 0.
+                    htlc_risk: self
+                        .htlcs
+                        .htlc_risk(forward.fee_msat(), forward.expiry_out_height),
+                },
+                outgoing_reputation: ReputationValues {
+                    reputation: outgoing_reputation,
+                    revenue_threshold: incoming_revenue_threshold,
+                    in_flight_total_risk: self.htlcs.channel_in_flight_risk(
+                        ChannelFilter::OutgoingChannel(forward.outgoing_channel_id),
+                    ),
+                    htlc_risk: self
+                        .htlcs
+                        .htlc_risk(forward.fee_msat(), forward.expiry_in_height),
+                },
             },
             // The incoming channel can only use congestion resources if it hasn't recently misused congestion
             // resources and it doesn't currently have any htlcs using them.
@@ -201,24 +223,36 @@ impl ForwardManagerImpl {
                         forward.outgoing_channel_id,
                         ResourceBucketType::General,
                     ),
-                    slots_available: outgoing_channel.general_bucket.slot_count,
+                    slots_available: outgoing_channel
+                        .outgoing_direction
+                        .general_bucket
+                        .slot_count,
                     liquidity_used_msat: self.htlcs.bucket_in_flight_msat(
                         forward.outgoing_channel_id,
                         ResourceBucketType::General,
                     ),
-                    liquidity_available_msat: outgoing_channel.general_bucket.liquidity_msat,
+                    liquidity_available_msat: outgoing_channel
+                        .outgoing_direction
+                        .general_bucket
+                        .liquidity_msat,
                 },
                 congestion_bucket: BucketResources {
                     slots_used: self.htlcs.bucket_in_flight_count(
                         forward.outgoing_channel_id,
                         ResourceBucketType::Congestion,
                     ),
-                    slots_available: outgoing_channel.congestion_bucket.slot_count,
+                    slots_available: outgoing_channel
+                        .outgoing_direction
+                        .congestion_bucket
+                        .slot_count,
                     liquidity_used_msat: self.htlcs.bucket_in_flight_msat(
                         forward.outgoing_channel_id,
                         ResourceBucketType::Congestion,
                     ),
-                    liquidity_available_msat: outgoing_channel.congestion_bucket.liquidity_msat,
+                    liquidity_available_msat: outgoing_channel
+                        .outgoing_direction
+                        .congestion_bucket
+                        .liquidity_msat,
                 },
             },
         })
@@ -427,6 +461,9 @@ impl ReputationManager for ForwardManager {
             reputations.insert(
                 *scid,
                 ChannelSnapshot {
+                    incoming_reputation: channel
+                        .incoming_direction
+                        .incoming_reputation(access_ins)?,
                     outgoing_reputation: channel
                         .outgoing_direction
                         .outgoing_reputation(access_ins)?,
--- a/ln-resource-mgr/src/lib.rs
+++ b/ln-resource-mgr/src/lib.rs
@@ -139,11 +139,11 @@ pub enum FailureReason {
     NoReputation,
 }
 
-/// A snapshot of the reputation and resources available for a forward.
+/// A snapshot of the incoming and outgoing reputation and resources available for a forward.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize)]
 pub struct AllocationCheck {
-    /// The reputation values used to compare the incoming channel's revenue to the outgoing channel's reputation for
-    /// the htlc proposed.
+    /// The reputation values used to check the incoming and outgoing reputation for the htlc
+    /// proposed.
     pub reputation_check: ReputationCheck,
     /// Indicates whether the incoming channel is eligible to consume congestion resources.
     pub congestion_eligible: bool,
@@ -190,7 +190,11 @@ impl AllocationCheck {
     ) -> Result<ResourceBucketType, FailureReason> {
         match incoming_endorsed {
             EndorsementSignal::Endorsed => {
-                if self.reputation_check.sufficient_reputation() {
+                if self
+                    .reputation_check
+                    .outgoing_reputation
+                    .sufficient_reputation()
+                {
                     Ok(ResourceBucketType::Protected)
                 } else {
                     // If the htlc was endorsed but the peer doesn't have reputation, we consider giving them a shot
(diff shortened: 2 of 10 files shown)


elnosh:
just noticed that with some of the recent changes I missed adding an `incoming_reputation` field to `ChannelSnapshot` when returning the list of channels from the forward manager.

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -276,18 +285,26 @@ impl ReputationManager for ForwardManager {
...
+                    liquidity_msat: congestion_liquidity_amount,
+                };
+
                 v.insert(TrackedChannel {
-                    incoming_direction: IncomingChannel::new(self.params.reputation_params),
+                    incoming_direction: IncomingChannel::new(
+                        self.params.reputation_params,
+                        general_bucket.clone(),
+                        congestion_bucket.clone(),
+                    ),
```
carlaKC, line 391: We don't need a bucket on the incoming channel, this is the one asymmetry that we have  in all this - we can look at reputation in the outgoing or incoming direction but we're only allocating resources on the outgoing side (since we've already accepted the incoming HTLC).

I don't think that this is clearly written down anywhere, and I probably didn't do a good job at explaining it - my bad!

ln-resource-mgr/src/htlc_manager.rs
```diff
@@ -145,26 +146,48 @@ impl InFlightManager {
...
-        outgoing_channel_id: u64,
+        channel: ChannelFilter,
         bucket: ResourceBucketType,
     ) -> u64 {
-        self.in_flight
-            .iter()
-            .filter(|(_, v)| v.bucket == bucket && v.outgoing_channel_id == outgoing_channel_id)
-            .map(|(_, v)| v.outgoing_amt_msat)
-            .sum()
+        match channel {
```
carlaKC, line 152: These can be _slightly_ abbreviated (likewise for `bucket_in_flight_count`):

```
self.in_flight.iter().filter_map(|(htlc_ref, in_flight)| {
		match filter {
			ChannelFilter::IncomingChannel(scid) => {
				if htlc_ref.channel_id == scid && in_flight.bucket == bucket {
					Some(in_flight.incoming_amt_msat)
				}else {
					None
				}
			},
			ChannelFilter::OutgoingChannel(scid) => {
				if in_flight.bucket == bucket && in_flight.outgoing_channel_id == scid {
					Some(in_flight.outgoing_amt_msat)
				}else {
					None
				}
			}
		}
	}).sum()
```
I'm not sure it's much better than what we have here so non-blocking.

(I should have used `filter_map` above as well, my oversight)

ln-resource-mgr/src/htlc_manager.rs
```diff
@@ -145,26 +146,48 @@ impl InFlightManager {
...
+                .in_flight
+                .iter()
+                .filter(|(htlc_ref, in_flight)| {
+                    htlc_ref.channel_id == scid && in_flight.bucket == bucket
+                })
+                .count() as u16,
+            ChannelFilter::OutgoingChannel(scid) => {
+                self.in_flight
+                    .iter()
+                    .filter(|(_, v)| v.bucket == bucket && v.outgoing_channel_id == scid)
```
carlaKC, line 187: nit: second value is `in_flight` above and `v` here

ln-resource-mgr/src/htlc_manager.rs
```diff
@@ -145,26 +146,48 @@ impl InFlightManager {
     /// Returns the total balance of htlcs in flight in the bucket provided.
     pub(super) fn bucket_in_flight_msat(
         &self,
-        outgoing_channel_id: u64,
+        channel: ChannelFilter,
```
carlaKC, line 149: nit: channel/filter to align with above impls

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -159,44 +159,146 @@ impl ForwardManagerImpl {
...
+            ReputationError::ErrOutgoingNotFound(forward.outgoing_channel_id),
+        )?;
 
-        Ok(AllocationCheck {
-            reputation_check: ReputationCheck {
-                outgoing_reputation: outgoing_channel.outgoing_reputation(forward.added_at)?,
-                revenue_threshold,
-                in_flight_total_risk: self.htlcs.channel_in_flight_risk(
+        // Revenue over the outgoing channel that the incoming channel must meet.
+        let outgoing_treshold = outgoing_channel
```
carlaKC, line 236: outgoing_treshold/outgoing_threshold

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -159,44 +159,146 @@ impl ForwardManagerImpl {
...
+            reputation_check: ReputationCheck {
+                incoming_reputation: ReputationValues {
+                    reputation: incoming_reputation,
+                    revenue_treshold: outgoing_treshold,
+                    in_flight_total_risk: self.htlcs.channel_in_flight_risk(
+                        ChannelFilter::IncomingChannel(forward.incoming_ref.channel_id),
+                    ),
+                    htlc_risk: self
+                        .htlcs
+                        .htlc_risk(forward.fee_msat(), forward.expiry_in_height),
```
carlaKC, line 295: For incoming reputation we want to look at the outgoing expiry height for htlc risk

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -159,44 +159,146 @@ impl ForwardManagerImpl {
...
+        let incoming_reputation = incoming_channel
+            .incoming_direction
+            .incoming_reputation(forward.added_at)?;
+
         let no_congestion_misuse = incoming_channel
             .incoming_direction
             .no_congestion_misuse(forward.added_at);
 
+        // Bucket resources available on the incoming channel.
+        let incoming_resources = ResourceValues {
```
carlaKC, line 191: Can we cleanly move creation of `ResourceValues` into a helper function? 

I think it should be doable by passing in a `incoming_reputation: bool` and `&TrackedChannel`?

ln-resource-mgr/src/lib.rs
```diff
@@ -250,39 +263,60 @@ impl AllocationCheck {
...
+    pub reputation: i64,
+    pub revenue_treshold: i64,
     pub in_flight_total_risk: u64,
     pub htlc_risk: u64,
 }
 
-impl ReputationCheck {
-    /// Returns a boolean indicating whether the outgoing channel has sufficient reputation for this htlc to be
-    /// forwarded to it.
+impl ReputationValues {
```
carlaKC, line 337: nice 👍 

nit: revenue_treshold typo in a few places

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -117,9 +117,17 @@ impl RevenueAverage {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Reputation {
+    Incoming,
+    Outgoing,
+    Both,
```
carlaKC, line 124: nitnit: Both/Bidirectional

ln-resource-mgr/src/lib.rs
```diff
@@ -207,12 +226,31 @@ impl AllocationCheck {
...
             }
             EndorsementSignal::Unendorsed => {
-                if self
-                    .resource_check
-                    .outgoing_resources
-                    .general_bucket
-                    .resources_available(htlc_amt_msat)
-                {
+                let sufficient_resources = match reputation_check {
+                    Reputation::Incoming => self
```
carlaKC, line 230: Can this be implemented on `Reputation` instead? Likewise for `sufficient_reputation`

carlaKC (commented):
Really nicely structured PR 🥇 

Just done a first pass on this, my only major comment is that we don't need any buckets for incoming (while we check reputation in both directions, we're only allocating resources on the outgoing channel). We also don't need a congestion bucket (I saw a todo about that).

Should simplify things a bit! Otherwise comments mostly aesthetic, around pulling some things into helpers.

Also feel free to just change commits + force push (rather than do fixups) is that suits you, happy to review either way since the diff isn't so big.

ln-resource-mgr/src/lib.rs
elnosh: this I could also probably delete and leave as it was before since no need to check bucket on incoming channel

ln-resource-mgr/src/forward_manager.rs
carlaKC: Right, we only need `ReputationValues` for reputation in both directions 👍

elnosh:
thanks for the clarification! removed all the bucketing for incoming channel and addressed other comments.

ln-resource-mgr/src/htlc_manager.rs
elnosh: changed this to use `filter_map` but didn't for `bucket_in_flight_count` because there we are just getting the count and not using `map` after `filter`

ln-resource-mgr/src/htlc_manager.rs
```diff
@@ -145,26 +146,53 @@ impl InFlightManager {
...
+        filter: ChannelFilter,
         bucket: ResourceBucketType,
     ) -> u16 {
-        self.in_flight
-            .iter()
-            .filter(|(_, v)| v.bucket == bucket && v.outgoing_channel_id == outgoing_channel_id)
-            .count() as u16 // Safe because we have in protocol limit 483.
+        match filter {
+            ChannelFilter::IncomingChannel(scid) => self
+                .in_flight
```
carlaKC, line 181: nit: could be de-duped as above, non-blocking

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -150,44 +150,66 @@ impl ForwardManagerImpl {
...
-            .ok_or(ReputationError::ErrOutgoingNotFound(
-                forward.outgoing_channel_id,
-            ))?
-            .outgoing_direction;
+        let outgoing_channel = self.channels.get_mut(&forward.outgoing_channel_id).ok_or(
+            ReputationError::ErrOutgoingNotFound(forward.outgoing_channel_id),
+        )?;
+
+        // Revenue over the outgoing channel that the incoming channel must meet.
+        let outgoing_threshold = outgoing_channel
```
carlaKC, line 212: `outgoing_revenue_threshold` to match `incoming_revenue_threshold` above?

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -150,44 +150,66 @@ impl ForwardManagerImpl {
...
+                },
+                outgoing_reputation: ReputationValues {
+                    reputation: outgoing_reputation,
+                    revenue_threshold: incoming_revenue_threshold,
+                    in_flight_total_risk: self.htlcs.channel_in_flight_risk(
+                        ChannelFilter::OutgoingChannel(forward.outgoing_channel_id),
+                    ),
+                    // The underlying simulation is block height agnostic, and starts its routes with a height of zero, so
+                    // we can just use the incoming expiry to reflect "maximum time htlc can be held on channel", because
+                    // we're calculating expiry_in_height - 0.
```
carlaKC, line 241: Let's move this comment up and generalize it to being able to use expiry height (since this applies to `expiry_out_height` above as well)

ln-simln-jamming/src/parsing.rs
```diff
@@ -128,6 +128,14 @@ pub struct Cli {
...
     #[arg(long)]
     pub attacker_alias: String,
+
+    /// Use incoming reputation for the simulation.
+    #[arg(long)]
+    pub incoming_reputation: bool,
+
+    /// Use outgoing reputation for the simulation.
+    #[arg(long)]
+    pub outgoing_reputation: bool,
```
carlaKC, line 138: `incoming/outgoing_reputation_only` to make a it a little clearer that we're restricting what we check?
could also be interpreted as needing both set for bidirectional, but actually you should set neither.

carlaKC (commented):
Looks great! Just nits left from me.

Need to do a closer pass on the tests + do some manual testing, then this is good to go!

ln-resource-mgr/src/forward_manager.rs
elnosh: oops, yeah thank you

ln-resource-mgr/src/htlc_manager.rs
elnosh: how? went through iterator methods and couldn't find a way to abbreviate it

elnosh:
> I missed adding an `incoming_reputation` field to `ChannelSnapshot`

added this but I didn't change `count_reputation_pairs` to include it since we are not changing the outcome of the simulation yet based on incoming reputation. Not sure if I should change that method here or will that be addressed later? Left an ugly comment in a test regarding this.

I also broke up the bootstrap method in `ReputationInterceptor` to make clippy happy but I think we'll get rid of this in the future if we start up the simulation from a reputation file rather than replaying all the forwards.

ln-resource-mgr/src/htlc_manager.rs
carlaKC: Doing the match on channel type inside of filter:
```
self.in_flight
            .iter()
            .filter(|(htlc_ref, in_flight)| match filter {
                ChannelFilter::IncomingChannel(scid) => {
                    htlc_ref.channel_id == scid && in_flight.bucket == bucket
                }
                ChannelFilter::OutgoingChannel(scid) => {
                    in_flight.bucket == bucket && in_flight.outgoing_channel_id == scid
                }
            })
            .count() as u16
```

ln-resource-mgr/src/htlc_manager.rs
```diff
@@ -145,26 +146,53 @@ impl InFlightManager {
     /// Returns the total balance of htlcs in flight in the bucket provided.
     pub(super) fn bucket_in_flight_msat(
         &self,
-        outgoing_channel_id: u64,
+        filter: ChannelFilter,
```
carlaKC, line 149: We don't actually need a filter for `bucket_in_flight_count` and `bucket_in_flight_msat`, since we only ever call this to get the bucket utilization on the outgoing link, right?

ln-resource-mgr/src/lib.rs
```diff
@@ -139,15 +139,15 @@ pub enum FailureReason {
...
 pub struct AllocationCheck {
-    /// The reputation values used to compare the incoming channel's revenue to the outgoing channel's reputation for
-    /// the htlc proposed.
+    /// The reputation values used to check the incoming and outgoing reputation for the htlc
+    /// proposed.
     pub reputation_check: ReputationCheck,
     /// Indicates whether the incoming channel is eligible to consume congestion resources.
     pub congestion_eligible: bool,
-    /// The resources available on the outgoing channel.
+    /// The bucket resources available on the incoming and outgoing channel.
```
carlaKC, line 151: nit: comment can be unchanged since we only give outgoing channel

ln-resource-mgr/src/lib.rs
```diff
@@ -510,6 +528,7 @@ mod tests {
                     liquidity_used_msat: 0,
                     liquidity_available_msat: 200_000,
                 },
+
```
carlaKC, line 531: nit: unnecessary newline

ln-simln-jamming/src/analysis.rs
```diff
@@ -57,17 +57,32 @@ impl Serialize for Record {
                 .forwarding_outcome(self.forward.amount_in_msat, self.forward.incoming_endorsed),
         )?;
         state.serialize_field(
-            "revenue_threshold",
-            &self.decision.reputation_check.revenue_threshold,
+            "incoming_revenue",
```
carlaKC, line 63: incoming_revenue/incoming_revenue_threshold + make `in_flight_risk` directional and write for both

I think we can also go ahead and add `outgoing_revenue_threshold` and `incoming_reputation`, all the values are useful here 👍

ln-resource-mgr/src/lib.rs
```diff
@@ -597,15 +597,49 @@ mod tests {
         let check = test_congestion_check();
         assert!(
             check
-                .inner_forwarding_outcome(10, EndorsementSignal::Endorsed)
+                .inner_forwarding_outcome(10, EndorsementSignal::Endorsed, Reputation::Incoming)
```
carlaKC, line 600: What's the rationale for changing the reputation scheme for each `inner_forwarding_outcome` call here?

afaik it's not relevant to what's being tested, which makes this a little less readable.

Same comment generally for a few of the tests, unless I'm missing something.

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -176,27 +176,17 @@ where
         })
     }
 
-    /// Creates a network from the set of edges provided and bootstraps the reputation of nodes in the network using
-    /// the historical forwards provided. Forwards are expected to be sorted by added_ns in ascending order.
-    pub async fn new_with_bootstrap(
-        params: ForwardManagerParams,
-        edges: &[NetworkParser],
-        general_jammed: &[(u64, PublicKey)],
+    /// Bootstraps the reputation of nodes in the interceptor network using the historical forwards provided.
```
carlaKC, line 179: Commit unrelated to incoming reputation?

ln-resource-mgr/src/lib.rs
```diff
@@ -187,14 +189,11 @@ impl AllocationCheck {
...
+        reputation_check: Reputation,
     ) -> Result<ResourceBucketType, FailureReason> {
         match incoming_endorsed {
             EndorsementSignal::Endorsed => {
-                if self
-                    .reputation_check
-                    .outgoing_reputation
-                    .sufficient_reputation()
-                {
+                if reputation_check.sufficient_reputation(self) {
```
carlaKC, line 196: We've got a slight asymmetry issue here with incoming vs outgoing reputation. 

This is all correct, except for the case where we have an endorsed HTLC that doesn't have sufficient reputation.

Bidirectional: we drop the HTLC with `NoResources`
Outgoing: we drop the HTLC with `NoResources`
Incoming: we forward the HTLC in the general bucket

Rationale being that if we're looking at reputation in the outgoing direction, we can't risk forwarding an endorsed htlc to an unknown peer (they may hold on to it and use it to try trash our reputation with our upstream peers).

Written down (poorly) here: https://gist.github.com/carlaKC/6762d88903d1cc27339859816ed80d43

ln-resource-mgr/src/htlc_manager.rs
elnosh: ah right, I changed this initially when I thought we needed bucketing from channels in both directions.

ln-simln-jamming/src/reputation_interceptor.rs
elnosh: kind of, I added the `reputation_check` param to the method and then clippy was complaining that it had too many. So I broke it up to bootstrap separately.

ln-resource-mgr/src/lib.rs
elnosh: hmmm, iiuc this test should return `ResourceBucketType::Congestion` if the reputation check fails so I added a case for each of the 3 reputation checks. Although if the congestion bucket only applies for outgoing channel, you mean it should only be useful in `Reputation::Outgoing`?

ln-resource-mgr/src/lib.rs
elnosh: > This is all correct, except for the case where we have an endorsed HTLC that doesn't have sufficient reputation.
> 
> Bidirectional: we drop the HTLC with `NoResources` 
> Outgoing: we drop the HTLC with `NoResources` 
> Incoming: we forward the HTLC in the general bucket

I'm not following, if the `sufficient_reputation` check fails, we are still dropping the htlc but why change the failure reason from the current `NoReputation` to `NoResources`? 

So if the the reputation type is `Incoming` and it does not have enough, default to general bucket?

ln-resource-mgr/src/htlc_manager.rs
elnosh: I think I can delete that entire commit where I changed those methods to handle `Incoming/Outgoing` since none of that is needed?

ln-resource-mgr/src/lib.rs
carlaKC: > I'm not following, if the sufficient_reputation check fails, we are still dropping the htlc but why change the failure reason from the current NoReputation to NoResources?

Sorry my bad, it stays `NoReputation`!

> So if the the reputation type is Incoming and it does not have enough, default to general bucket?

Yeah exactly.

ln-resource-mgr/src/htlc_manager.rs
carlaKC: Yeah we can drop 8369ecc23bc4c3409383f0ba2cc43f5d9ec7f351 👍

ln-resource-mgr/src/lib.rs
carlaKC: Right, gotcha! perhaps we could add a test helper here and pass each reputation scheme in, something like:
`test_inner_forwarding_outcome_congestion_for_reptuation(scheme: Reputation) {}`

Then call each one in `test_inner_forwarding_outcome_congestion`:
```
test_inner_forwarding_outcome_congestion_for_reputation(Reputation::Incoming);
test_inner_forwarding_outcome_congestion_for_reputation(Reputation::Outgoing);
test_inner_forwarding_outcome_congestion_for_reputation(Reputation::Bidirectional);
```

ln-resource-mgr/src/lib.rs
```diff
@@ -260,27 +264,35 @@ impl AllocationCheck {
...
 }
 
-/// A snapshot of a reputation check for a htlc forward.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize)]
 pub struct ReputationCheck {
-    pub outgoing_reputation: i64,
+    /// Reputation values of the incoming channel for the proposed htlc.
+    pub incoming_reputation: ReputationValues,
+    /// Reputation values of the outgoing channel for the proposed htlc.
+    pub outgoing_reputation: ReputationValues,
```
carlaKC, line 284: nit: the values in `incoming_reputation` actually look at the incoming channel's reputation and the outgoing channel's revenue, so "of the incoming channel" doesn't really parse here (same for outgoing).

Perhaps: The values used to check incoming/outgoing reputation for the channel pair.

ln-simln-jamming/src/analysis.rs
```diff
@@ -57,17 +57,60 @@ impl Serialize for Record {
                 .forwarding_outcome(self.forward.amount_in_msat, self.forward.incoming_endorsed),
         )?;
         state.serialize_field(
-            "revenue_threshold",
-            &self.decision.reputation_check.revenue_threshold,
+            "incoming_revenue_threshold",
```
carlaKC, line 65: WDYT about naming these by the reputation check we're doing rather than the directions?

```
rep_outgoing_threshold = outgoing_reputation.reputation
rep_outgoing_reputation = outgoing_reputation.threshold
rep_outgoing_risk = outgoing_reputation.htlc_risk

rep_incoming_threshold = incoming_reputation.reputation
rep_incoming_reputation = incoming_reputation.threshold
rep_incoming_risk = incoming_reputation.htlc_risk
```

I think that this is how you'd look at the data (checking one direction of reputation check at a time), and IMO it makes it a little easier to group these values.

ln-resource-mgr/src/lib.rs
```diff
@@ -203,7 +202,20 @@ impl AllocationCheck {
                         return Ok(ResourceBucketType::Congestion);
                     }
 
-                    Err(FailureReason::NoReputation)
+                    match reputation_check {
+                        Reputation::Incoming => {
```
carlaKC, line 206: nit: could be `if reputation_check == Reputation::Incoming` to save on some match boilerplate, but no strong feelings here.

ln-resource-mgr/src/lib.rs
```diff
@@ -203,7 +202,20 @@ impl AllocationCheck {
                         return Ok(ResourceBucketType::Congestion);
                     }
 
-                    Err(FailureReason::NoReputation)
+                    match reputation_check {
```
carlaKC, line 205: Let's add a comment here explaining that if we've got outgoing reputation in play, we're required to drop endorsed htlcs to peers without reputation to protect against the unknown peer possibly damaging our reputation with our upstream peers + if we're only looking at incoming, we're not at risk because we're not held accountable for the behavior of downstream nodes.

carlaKC (commented):
Looking great! Just needs a comment on the new reputation logic + possibly a second look at how we're naming the values we write to CSV for incoming/outgoing reputation (open to discussion, this is always sort of mind boggling to think about!).

ln-resource-mgr/src/lib.rs
elnosh: I slightly prefer match over if let here as I feel it's bit more readable so I left as is. But I'm not hard set either. If you want, I can change it.

elnosh:
added comment on the new reputation logic and renamed fields as suggested to group them by incoming/outgoing reputation.

PR #41, opened 2025-04-15
Make attack generic in simulator

carlaKC:
Depends on #37, WIP at pulling attacks out into a trait

### DIFF
--- a/ln-simln-jamming/src/attacks/mod.rs
+++ b/ln-simln-jamming/src/attacks/mod.rs
@@ -1,0 +1,59 @@
+use async_trait::async_trait;
+use bitcoin::secp256k1::PublicKey;
+use simln_lib::sim_node::InterceptRequest;
+
+use crate::{endorsement_from_records, records_from_endorsement, BoxError, NetworkReputation};
+
+#[allow(dead_code)]
+pub struct NetworkSetup {
+    /// The identifier for channel edges that should be general jammed during the simulation.
+    ///
+    /// For example: a channel with ID 999 between A -- B will have general resources exhausted as follows:
+    /// - (999, A): no general resources for A -> B
+    /// - (999, B): no general resources for B -> A
+    ///
+    /// This option is provided as a convenience for attacks that don't wish to implement general jamming the cost of
+    /// this general jamming will be accounted for at the end of the attack.
+    pub general_jammed_nodes: Vec<(u64, PublicKey)>,
+}
+
+// Defines an attack that can be mounted against the simulation framework.
+#[async_trait]
+#[allow(dead_code)]
+pub trait JammingAttack {
+    /// Responsible for validating that the network provided meets any topological expectations for the attack, and
+    /// returning network-specific setup instructions for the attack.
+    ///
+    /// The default implementation has no network setup and passes validation.
+    fn setup_for_network(&self) -> Result<NetworkSetup, BoxError> {
+        Ok(NetworkSetup {
+            general_jammed_nodes: vec![],
+        })
+    }
+
+    /// Called for evey HTLC that is forwarded through or to attacking nodes, to allow the attacker to take custom
+    /// actions on HTLCs. This function may block, as it is spawned in a task, but *must* eventually send a response to
+    /// the request.
+    ///
+    /// The default implementation will forward HTLCs immediately, copying whatever incoming endorsement signal it
+    /// received.
+    async fn intercept_attacker_htlc(&self, req: InterceptRequest) -> Result<(), BoxError> {
+        req.response
+            .send(Ok(Ok(records_from_endorsement(endorsement_from_records(
+                &req.incoming_custom_records,
+            )))))
+            .await
+            .map_err(|e| e.into())
+    }
+
+    /// Returns a boolean that indicates whether a shutdown condition for the simulation has been reached.
+    ///
+    /// Should be used when there are shutdown conditions specific to the attack, the default implementation will
+    /// return `Ok(false)`.
+    async fn simulation_completed(
+        &self,
+        _start_reputation: NetworkReputation,
+    ) -> Result<bool, BoxError> {
+        Ok(false)
+    }
+}
--- a/ln-simln-jamming/src/attacks/mod.rs
+++ b/ln-simln-jamming/src/attacks/mod.rs
@@ -4,6 +4,8 @@ use simln_lib::sim_node::InterceptRequest;
 
 use crate::{endorsement_from_records, records_from_endorsement, BoxError, NetworkReputation};
 
+pub mod sink;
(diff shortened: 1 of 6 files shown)


ln-simln-jamming/src/attacks/sink.rs
```diff
@@ -0,0 +1,516 @@
...
+    start_reputation: &NetworkReputation,
+    current_reputation: &NetworkReputation,
+) -> Result<bool, BoxError> {
+    if current_reputation.attacker_reputation == 0 {
+        log::error!("Attacker has no more reputation with the target");
+
+        if current_reputation.target_reputation >= start_reputation.target_reputation {
+            log::error!("Attacker has no more reputation with target and the target's reputation is similar to simulation start");
+            return Ok(true);
+        }
```
elnosh, line 157: was this separated into it's own function for better testing? trying to understand why not have this logic in `simulation_completed`

ln-simln-jamming/src/attacks/sink.rs
```diff
@@ -0,0 +1,516 @@
...
+        );
+
+        // If the htlc is endorsed, then we go ahead and hold the htlc for as long as we can only exiting if we
+        // get a shutdown signal elsewhere.
+        let resp = select! {
+            _ = self.listener.clone() => Err(ForwardingError::InterceptorError("shutdown signal received".to_string().into())),
+            _ = self.clock.sleep(max_hold_secs) => Ok(records_from_endorsement(EndorsementSignal::Endorsed))
+        };
+
+        send_intercept_result!(req, Ok(resp), self.shutdown);
```
elnosh, line 143: related to a previous comment about blocking, this may not be strictly related to this PR since it already works like this but the hodling here could be done in the background and communication of result could be done in the background so that the function returns immediately?

elnosh (commented):
overall LGTM, just have some questions for me to understand a couple things better (:

ln-simln-jamming/src/attacks/sink.rs
carlaKC: Yep - I'll add a comment here! Bit of a mare to construct a mock of the return for `get_network_reputation` that will return the specific `NetworkReputation` values that we want to test. We already have separate coverage that does that and it's pretty verbose, so I thought it's okay to just skip over that part in these tests.

ln-simln-jamming/src/attacks/sink.rs
carlaKC: > hodling here could be done in the background and communication of result could be done in the background so that the function returns immediately?

Definitely, I just don't see what we gain here by spinning up a task to deliver this result after the function has returned?

The interceptor on the sim-ln side was specifically designed to allow interceptor functions that block for a while so that implementations can be a bit lazier.

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -100,19 +100,117 @@ where
...
 /// network.
-#[derive(Clone)]
 pub struct ReputationInterceptor<R, M>
 where
     R: ForwardReporter,
     M: ReputationManager,
 {
-    network_nodes: Arc<Mutex<HashMap<PublicKey, Node<M>>>>,
+    inner: Mutex<InnerReputationInterceptor<R, M>>,
+}
```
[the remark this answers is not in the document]
carlaKC: I added this commit when I was in absolute `async_trait` + generics hell, possibly unnecessary or has an outdated comment - I'll take another look at it 👍 

More generally I've been borrowing this pattern from LDK, nice way to handle locking when it doesn't need to be super granular + avoids a few `&mut self` function signatures.

ln-simln-jamming/src/attacks/sink.rs
elnosh: > The interceptor on the sim-ln side was specifically designed to allow interceptor functions that block for a while so that implementations can be a bit lazier.

then you can ignore my comment hehe. It's just that from looking at the interceptor in sim-ln, it seemed to me that the intent was for it to be non-blocking since the result of the intercepted htlc is sent through a channel.

ln-simln-jamming/src/attacks/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+        })
+    }
+
+    /// Called for evey HTLC that is forwarded through or to attacking nodes, to allow the attacker to take custom
+    /// actions on HTLCs. This function may block, as it is spawned in a task, but *must* eventually send a response to
+    /// the request.
+    ///
+    /// The default implementation will forward HTLCs immediately, copying whatever incoming endorsement signal it
+    /// received.
+    async fn intercept_attacker_htlc(&self, req: InterceptRequest) -> Result<(), BoxError> {
```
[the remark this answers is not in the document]
elnosh: tied to comment below https://github.com/carlaKC/jam-ln/pull/41#discussion_r2060960713 if the intent with `intercept_htlc` is to be blocking, then you can ignore

ln-simln-jamming/src/attacks/sink.rs
carlaKC: Looking at this again, I'm wrong. If we happen to hit this long-held [intercept](https://github.com/carlaKC/sim-ln/blob/interceptor-latency/simln-lib/src/sim_node.rs#L1256) first, and we have multiple interceptors, the other ones will have to wait until the hold is complete before it's reported to the rest of them (if the long-held one is hit first). 

This isn't a disaster (we [drop](https://github.com/carlaKC/sim-ln/blob/interceptor-latency/simln-lib/src/sim_node.rs#L1229) our top level lock so that the task that's waiting on all the interceptors isn't holding up anything else in the simulator), but it is annoying.

Especially if you have one interceptor that holds for long, and one that immediately returns a fail because there's an unnecessary wait when the htlc is just going to be failed at the end of it anyway.

Feels like something we should fix on the sim-ln side rather than here (because each interceptor impl will need to worry about this). Maybe we just need to spin up the `intercept_htlc` functions in their own tasks + await them rather than having channels - that way everything can block and we can collect results one by one.

For this PR, going to leave as-is since it's just a code move.

ln-simln-jamming/src/reputation_interceptor.rs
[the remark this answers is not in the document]
carlaKC: Ok yup this commit is completely unnecessary :')

Forward manager was a little different, because we actually did care about locking more there - locking channels and htlcs separately could lead to inconsistency so the high level lock [fixed the issue of needing to acquire 2x locks](https://github.com/carlaKC/jam-ln/pull/28/commits/bd45732ddc113798c3c11221c8528e9d7cbc9dc1) for forwarding checks.

PR #42, opened 2025-04-16
improve startup: start from snapshot (2/2)

elnosh:
2nd part of https://github.com/carlaKC/jam-ln/issues/31. Builds upon https://github.com/carlaKC/jam-ln/pull/38.

This will start a `ReputationInterceptor` from a previously generated file containing the reputation snapshot for the network instead of building it by replaying all the forwards from a bootstrap file. 

It then reads the peacetime file as how it is doing already and thens starts the simulation.


ln-simln-jamming/src/main.rs
```diff
@@ -183,18 +166,29 @@ async fn main() -> Result<(), BoxError> {
...
 
+    let reputation_snapshot = reputation_snapshot_from_file(&cli.reputation_file)?;
+    let target_channels_snapshot = reputation_snapshot
+        .get(&target_pubkey)
+        .ok_or::<BoxError>("target pubkey not found".to_string().into())?;
+
+    let bootstrap_revenue = target_channels_snapshot
+        .iter()
+        .map(|(_, snapshot)| snapshot.bidirectional_revenue)
+        .sum::<i64>() as u64;
```
carlaKC, line 177: Ooof didn't think about this when we first discussed this :(

The `bidirectional_revenue` that we track here is decayed over time, and the `bootstrap_revenue` that we need is the undecayed sum (so that we can compare to the peacetime projection).

We've got this available when we're generating the snapshot file, but isn't kept around after that.

Some approaches we could take are:
- Write another file with the target node's revenue + read that in on run
- Just print the revenue when we run the file generation and required running with this value as a param (seems footgunny)

Not in love with either of these options, open to other suggestions!

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -277,6 +278,21 @@ impl ReputationManager for ForwardManager {
...
+                let reputation = channel_reputation
+                    .as_ref()
+                    .map(|channel| channel.outgoing_reputation);
+
+                let revenue = match &channel_reputation {
+                    Some(channel) => {
+                        let mut revenue =
+                            RevenueAverage::new(&self.params.reputation_params, add_ins);
+                        // NOTE: not sure if Instant::now() here is ok?
+                        revenue.add_value(channel.bidirectional_revenue, Instant::now())?;
```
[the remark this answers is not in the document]
elnosh: > Perhaps we could pas in `channel_reputation: Option<ChannelSnapshot, Instant>` so that the caller can fix the "start" instant for all of these bootstrapped channels?

I'm a bit confused, how will that be different from the current `add_ins` that is currently being passed to `add_channel`?

ln-resource-mgr/src/outgoing_channel.rs
```diff
@@ -37,16 +37,27 @@ impl OutgoingChannel {
         params: ReputationParams,
         general_bucket: BucketParameters,
         congestion_bucket: BucketParameters,
+        outgoing_reputation: Option<i64>,
```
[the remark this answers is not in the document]
elnosh: agree, will prob need it in `ChannelSnapshot`

ln-simln-jamming/src/main.rs
elnosh: oh, right I didn't take that into account :man_facepalming:. I'd go writing it to another file

ln-resource-mgr/src/forward_manager.rs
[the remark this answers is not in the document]
carlaKC: Right, forgot that we already have `added_ins`!

My comment is that we shouldn't use `Instant::now()` because that's using regular wall clock time (and we're using a simulated/ sped up clock). But yes, `added_ins` is great, because it comes from our clock 👍

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -178,15 +179,79 @@ where
...
+                            clock.now(),
+                            Some(channel_snapshot.clone()),
+                        )?;
+                    }
+                }
+            }
+        }
 
-        // After the network has been bootstrapped, we can go ahead and general jam required channels.
+        for channel in edges {
```
carlaKC, line 219: Shouldn't `reputation_snapshot` have all the `edges` in our graph? 

Even if they didn't forward anything in the bootstrap period, they're still added to the reputation manager and written (even if reputation is 0)?

Would make sense to me to do a sanity check of bootstrap vs `edges`, making sure:
- They have the exact same set of channels/nodes
- They always have the same capacity 

Just to prevent any wrong file footguns

ln-simln-jamming/src/main.rs
```diff
@@ -156,14 +133,26 @@ async fn main() -> Result<(), BoxError> {
...
-    let reputation_interceptor = Arc::new(Mutex::new(ReputationInterceptor::new_for_network(
-        forward_params,
-        &sim_network,
-        reputation_check,
-        clock.clone(),
-        Some(results_writer),
-        shutdown.clone(),
-    )?));
+    let reputation_snapshot =
+        reputation_snapshot_from_file(&cli.reputation_dir.join("reputation-snapshot.csv"))?;
```
carlaKC, line 137: Nit: let's pull these out into constants and use them across read/write code?

carlaKC (commented):
Lookin good - just one q!

ln-simln-jamming/src/reputation_interceptor.rs
elnosh: ohh yeah.

> Would make sense to me to do a sanity check of bootstrap vs `edges`, making sure:
> 
>     * They have the exact same set of channels/nodes
> 
>     * They always have the same capacity

sounds good, will add that

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -178,15 +179,96 @@ where
...
+                            Some(channel_snapshot.clone()),
+                        )?;
+                        let alias = find_alias_by_pubkey(node_pubkey, edges)?;
+                        e.insert(Node::new(forward_manager, alias));
+                    }
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().forward_manager.add_channel(
+                            *scid,
+                            channel_snapshot.capacity_msat,
+                            clock.now(),
```
carlaKC, line 211: Let's use a single timestamp for all of the channels we add? Otherwise they're all added at different times, and will have their decaying averages adjusted differently to reflect the "present".

Probably best to pass this in as a parameter so the caller can control it.

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -178,15 +179,96 @@ where
...
+        edges: &[NetworkParser],
+        reputation_check: Reputation,
+        reputation_snapshot: HashMap<PublicKey, HashMap<u64, ChannelSnapshot>>,
+        clock: Arc<dyn InstantClock + Send + Sync>,
+        results: Option<Arc<Mutex<R>>>,
+        shutdown: Trigger,
+    ) -> Result<Self, BoxError> {
+        let mut network_nodes = HashMap::with_capacity(reputation_snapshot.len());
+
+        for (node_pubkey, channels) in reputation_snapshot.iter() {
```
carlaKC, line 193: I think that a more natural way to do this would be to loop through `edges` and then look up corresponding `reputation_snapshot` values (since that's where we have the hashmap.

This is also nice because we don't have to `find_alias_by_pubkey` - we have it on hand in `edges`.

Wanted to make sure this is doable myself, tried it with a macro:
```
        macro_rules! add_node_to_network {
            ($channel:expr, $node:tt) => {{
                let scid = $channel.scid.into();
                let pubkey = $channel.$node.pubkey;
                let alias = $channel.$node.alias.clone();

                let snapshot = reputation_snapshot
                    .get(&pubkey)
                    .ok_or(format!("node: {} not found in snapshot", pubkey))?
                    .get(&scid)
                    .ok_or(format!(
                        "channel: {} not found in snapshot for node: {}",
                        scid, pubkey
                    ))?;

                if snapshot.capacity_msat != $channel.capacity_msat {
                    return Err(format!(
                        "Channel {} has different capacities {} - {}",
                        scid, snapshot.capacity_msat, $channel.capacity_msat
                    )
                    .into());
                }

                match network_nodes.entry(pubkey) {
                    Entry::Vacant(e) => {
                        let forward_manager = ForwardManager::new(params);
                        forward_manager.add_channel(
                            $channel.scid.into(),
                            $channel.capacity_msat,
                            now,
                            Some(snapshot.clone()),
                        )?;
                        e.insert(Node::new(forward_manager, alias));
                    }
                    Entry::Occupied(mut e) => {
                        e.get_mut().forward_manager.add_channel(
                            $channel.scid.into(),
                            $channel.capacity_msat,
                            now,
                            Some(snapshot.clone()),
                        )?;
                    }
                }
            }};
        }

        for channel in edges {
			add_node_to_network!(channel, node_1); 
			add_node_to_network!(channel, node_2); 
        }
```

Once we've done this, the only other validation we'd need to do is make sure there are no entries in `reputation_snapshot` that _aren't_ in `channels`. I think that this can be done by just checking lengths (as `reputation_snapshot` should be 2x `edges` - each channel listed once per node):
```
        if edges.len() * 2
            != reputation_snapshot
                .values()
                .map(|chan_map| chan_map.len())
                .sum()
        {
            return Err("nice string here".into());
        }
```

ln-simln-jamming/src/main.rs
```diff
@@ -156,14 +134,24 @@ async fn main() -> Result<(), BoxError> {
...
-        reputation_check,
-        clock.clone(),
-        Some(results_writer),
-        shutdown.clone(),
-    )?));
+    let reputation_snapshot =
+        reputation_snapshot_from_file(&cli.reputation_dir.join(DEFAULT_REPUTATION_FILENAME))?;
+
+    let bootstrap_revenue: u64 =
+        std::fs::read_to_string(cli.reputation_dir.join(DEFAULT_REVENUE_FILENAME))?.parse()?;
```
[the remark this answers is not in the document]
elnosh: ah thanks. When running I was passing that arg rather than `join` with duration.

ln-simln-jamming/src/reputation_interceptor.rs
elnosh: although I've been somewhat reluctant to macros, yeah adding the channels and validation like this is better (:

ln-simln-jamming/src/reputation_interceptor.rs
elnosh: done along with macro change to add channels. Although still using the `clock` passed since that is still needed to init the interceptor

ln-simln-jamming/src/main.rs
```diff
@@ -188,20 +177,21 @@ async fn main() -> Result<(), BoxError> {
...
         target_channels.iter().map(|(k, v)| (*k, v.0)).collect();
+
     let start_reputation = get_network_reputation(
         reputation_interceptor.clone(),
         target_pubkey,
         attacker_pubkey,
         &target_pubkey_map,
         risk_margin,
-        now,
+        InstantClock::now(&*clock),
```
carlaKC, line 194: Do we need to change this?

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -178,15 +178,88 @@ where
...
     }
 
-    /// Bootstraps the reputation of nodes in the interceptor network using the historical forwards provided.
-    pub async fn bootstrap_network(
-        &mut self,
-        bootstrap: &BoostrapRecords,
-        general_jammed: &[(u64, PublicKey)],
-    ) -> Result<(), BoxError> {
-        self.bootstrap_network_history(bootstrap).await?;
+    pub async fn new_from_snapshot(
```
carlaKC, line 181: Tests to cover validation cases and a very basic snapshot being loaded + checking that we're adding the right values for the network.

ln-simln-jamming/src/main.rs
```diff
@@ -156,14 +134,25 @@ async fn main() -> Result<(), BoxError> {
...
-        clock.clone(),
-        Some(results_writer),
-        shutdown.clone(),
-    )?));
+    let reputation_dir = &cli.reputation_dir.join(cli.attacker_bootstrap.0.clone());
+    let reputation_snapshot =
+        reputation_snapshot_from_file(&reputation_dir.join(DEFAULT_REPUTATION_FILENAME))?;
+
+    let bootstrap_revenue: u64 =
+        std::fs::read_to_string(reputation_dir.join(DEFAULT_REVENUE_FILENAME))?.parse()?;
```
carlaKC, line 142: Previous comment has gotten lost - thoughts on also adding a sub-dir if we have `--attacker-bootstrap` set?

carlaKC (commented):
Looking good, would just like to get some test coverage on `new_from_snapshot` then I think this is good to go. 

Feel free to squash in the meantime 👍

ln-simln-jamming/src/main.rs
elnosh: hmmm yeah I changed it because I was getting `ErrUpdateInPast` here.

ln-simln-jamming/src/main.rs
elnosh: hmm not sure what you mean, I added the `reputation_dir` to use the `attacker-bootstrap` https://github.com/elnosh/jam-ln/blob/b804849b508b44a8dc659489ed1f78381e923fcb/ln-simln-jamming/src/main.rs#L137-L142

carlaKC: My bad - missed that line!

ln-simln-jamming/src/main.rs
carlaKC: Ok yeah that makes sense, we want to fix time up until the point where we create our `reputatoin_interceptor` but after that the clock has been set on the decaying averages so we need to start using the clock.

Let's add a comment here to make a note of that?

ln-simln-jamming/src/main.rs
```diff
@@ -156,14 +134,25 @@ async fn main() -> Result<(), BoxError> {
...
 
-    let reputation_interceptor = Arc::new(Mutex::new(ReputationInterceptor::new_for_network(
-        forward_params,
-        &sim_network,
-        reputation_check,
-        clock.clone(),
-        Some(results_writer),
-        shutdown.clone(),
-    )?));
+    let reputation_dir = &cli.reputation_dir.join(cli.attacker_bootstrap.0.clone());
```
carlaKC, line 137: Not for this PR, but we've got a slight inconsistency between write and read of these snapshots. 
`attacker_bootstrap` is optional for write but required for read, so we can generate a file that has no attacker bootstrap but can't run with it (becuase we require the CI flag). 

I think we can rela that requirement in a follow up?

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -1072,4 +1101,185 @@ mod tests {
             ForwardingError::InterceptorError(_),
         ));
     }
+
+    /// Tests starting interceptor from valid snapshot.
```
carlaKC, line 1105: Thanks for adding these - great to have the coverage!

carlaKC (approved):
Awesome change, thanks for going so many rounds on it!

Go ahead and squash then this is good to go!

Issue #44, opened 2025-04-29
Implement Upgradable Endorsement


carlaKC:
The simulator does not properly deal with endorsement at present because we can't set endorsement in `sim-ln`. 
It currently implements a half-baked version of the concept of [upgradable endorsement](https://gist.github.com/carlaKC/5139adf4fd12b4ecd53c660b5be11bf0#upgradeable-endorsement), which we should implement properly.

---

* Update `sim-ln` branch to accept a default set of `outgoing_custom_records` that it sets on the first hop of every payment
* Update `jamln` to set these as follows:
  * `ENDORSED_TYPE`: `106823` = 0
  * `UPGRADABLE_TYPE`: `106825` = 1
* Update `inner_forwarding_outcome` to opportunistically upgrade a HTLC's endorsement if:
  * `UPGRADABLE_TYPE=1`
  * `ENDORSED_TYPE=0`
  * The peer has sufficient reputation for the HTLC

In real life, the `UPGRADABLE_TYPE` will be provided in the onion so that it cannot be modified by malicious peers along the route. We enforce the following:
* If `UPGRADABLE_TYPE=0` and we receive a HTLC with `ENDORSED_TYPE=1`, fail it back; somebody has modified it

Bonus points: check that the output of `intercept_attacker_htlc` doesn't modify this TLV (buggy attack)

PR #46, opened 2025-05-01
implement upgradable endorsement

elnosh:
implements upgradable endorsements as described in https://github.com/carlaKC/jam-ln/issues/44

needs: https://github.com/carlaKC/sim-ln/pull/4

### DIFF
--- a/ln-resource-mgr/src/lib.rs
+++ b/ln-resource-mgr/src/lib.rs
@@ -138,6 +138,8 @@ pub enum FailureReason {
     NoResources,
     /// The outgoing peer has insufficient reputation for the htlc to occupy protected resources.
     NoReputation,
+    /// The upgradable signal has been tampered with so we should fail back the htlc.
+    UpgradableSignalModified,
 }
 
 /// A snapshot of the incoming and outgoing reputation and resources available for a forward.
@@ -166,9 +168,15 @@ impl AllocationCheck {
         &self,
         htlc_amt_msat: u64,
         incoming_endorsed: EndorsementSignal,
+        incoming_upgradable: bool,
         reputation_check: Reputation,
     ) -> ForwardingOutcome {
-        match self.inner_forwarding_outcome(htlc_amt_msat, incoming_endorsed, reputation_check) {
+        match self.inner_forwarding_outcome(
+            htlc_amt_msat,
+            incoming_endorsed,
+            incoming_upgradable,
+            reputation_check,
+        ) {
             Ok(bucket) => match bucket {
                 ResourceBucketType::General => {
                     ForwardingOutcome::Forward(EndorsementSignal::Unendorsed)
@@ -189,8 +197,13 @@ impl AllocationCheck {
         &self,
         htlc_amt_msat: u64,
         incoming_endorsed: EndorsementSignal,
+        incoming_upgradable: bool,
         reputation_check: Reputation,
     ) -> Result<ResourceBucketType, FailureReason> {
+        if !incoming_upgradable && incoming_endorsed == EndorsementSignal::Endorsed {
+            return Err(FailureReason::UpgradableSignalModified);
+        }
+
         match incoming_endorsed {
             EndorsementSignal::Endorsed => {
                 if reputation_check.sufficient_reputation(self) {
@@ -224,6 +237,10 @@ impl AllocationCheck {
                 }
             }
             EndorsementSignal::Unendorsed => {
+                if reputation_check.sufficient_reputation(self) && incoming_upgradable {
+                    return Ok(ResourceBucketType::Protected);
+                }
+
                 if self
                     .resource_check
                     .general_bucket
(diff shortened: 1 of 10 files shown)


ln-resource-mgr/src/lib.rs
```diff
@@ -138,6 +138,8 @@ pub enum FailureReason {
     NoResources,
     /// The outgoing peer has insufficient reputation for the htlc to occupy protected resources.
     NoReputation,
+    /// Upgradable signal has been modified.
```
carlaKC, line 141: nit: note that this means the htlc has been tampered with and should be failed back.

ln-simln-jamming/src/attack_interceptor.rs
```diff
@@ -38,149 +28,32 @@ where
     shutdown: Trigger,
 }
 
-impl<C, R, A> AttackInterceptor<C, R, A>
+impl<R, A> AttackInterceptor<R, A>
```
carlaKC, line 31: much code deleted, such good 😍

ln-simln-jamming/src/attack_interceptor.rs
```diff
@@ -221,8 +79,8 @@ where
                 records_from_endorsement(EndorsementSignal::Endorsed);
         }
 
-        // The target is not involved in the forward at all, just use jamming interceptor to implement reputation
-        // and bucketing.
+        // If attacker is not involved, use jamming interceptor to implement reputation and
+        // bucketing.
```
carlaKC, line 83: We can now remove the hard-set of `incoming_custom_records` - previously we were relying on `intercept_peer_outgoing` to filter out anything that wasn't upgradable and then setting endorsed here (because we know it must be upgradable).

This can now purely be handled by the `reputation_interceptor`! (and we get to delete moar code 😻 )

ln-simln-jamming/src/lib.rs
```diff
@@ -38,12 +42,22 @@ pub fn endorsement_from_records(records: &CustomRecords) -> EndorsementSignal {
...
+    match records.get(&UPGRADABLE_TYPE) {
+        Some(upgradable) => upgradable.len() == 1 && upgradable[0] == 1,
+        None => false,
+    }
+}
+
 /// Converts an endorsement signal to custom records using the blip-04 experimental TLV.
 pub fn records_from_endorsement(endorsement: EndorsementSignal) -> CustomRecords {
+    let mut records = CustomRecords::default();
+    records.insert(UPGRADABLE_TYPE, vec![1]);
```
carlaKC, line 57: nit: update docs to note that this will hard-set the `UPGRADABLE_TYPE`? 

Works for now because we always want this to be set, but maybe in the future we want to have some values that aren't upgradable.

ln-simln-jamming/src/lib.rs
```diff
@@ -38,12 +42,22 @@ pub fn endorsement_from_records(records: &CustomRecords) -> EndorsementSignal {
     }
 }
 
+/// Converts a set of custom tlv records to a bool signaling if the endorsement signal is
+/// upgradable.
+pub fn upgradable_from_records(records: &CustomRecords) -> bool {
```
carlaKC, line 47: WDYT about combining this with `endorsement_from_records`? `-> (EndorsementSignal, bool)`

I'm on the fence because we don't always need upgradable.

ln-simln-jamming/src/reputation_interceptor.rs
```diff
@@ -705,6 +711,29 @@ mod tests {
...
+            .await
+            .get_mut(&pubkeys[0])
+            .unwrap()
+            .forward_manager
+            .expect_add_htlc()
+            .return_once(|_| Ok(test_allocation_check(true)));
+
+        interceptor.intercept_htlc(request).await;
+
+        // should call add_htlc + return a reputation check that passes
```
carlaKC, line 732: nit: capitalize and end with `.`

carlaKC (commented):
Such a nice clean up of some ugly ad-hoccy code!

ln-simln-jamming/src/lib.rs
elnosh: I thought about doing it like that initially but ended up opting for the new method. Don't feel strongly about either. Mostly did it because it felt that this was more readable than the tuple:

` upgradable_endorsement: upgradable_from_records(&req.incoming_custom_records)`

elnosh:
> lgtm! Just needs dependencies bumped to now-merged branch 🦖

done! 

> But we can get rid of setting the incoming endorsement signal in the tests (since that's now done in reputation interceptor).

although I didn't change this because we are setting the endorsement signal through `setup_test_request` from test utils which we use in other places. 


I am perplexed by clippy. It's failing with 

```
error: unused import: `HashSet`
  --> ln-simln-jamming/src/main.rs:26:33
   |
26 | use std::collections::{HashMap, HashSet};
```

but it's used here: https://github.com/elnosh/jam-ln/blob/d81fc4e0c2384a088bea2df613030f58b8f9450d/ln-simln-jamming/src/main.rs#L89

elnosh:
whoops had to rebase

PR #47, opened 2025-05-06
makefile: install reputation_builder in install-tools command

carlaKC:
Add to makefile in a separate command. In future I can see us adding additional tooling binaries (such as one to generate the bootstrap data), so putting this under a generic title in preparation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,5 +12,8 @@ stable-output:
 
 check: check-code stable-output
 
+install-tools:
+	cargo install --locked --path ln-simln-jamming --bin reputation-builder
+
 install:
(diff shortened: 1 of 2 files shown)


elnosh (approved):
a not so useful comment for later - to make the executable be `reputation-builder` instead of with `_` will need to put it in a directory named like that?

carlaKC:
How about adding a name in [cargo.toml](https://github.com/carlaKC/jam-ln/compare/1a827456e02a80a71911542d9258d29bc2c8daad..49dd1caac18e4b04a025adca05af15976c15ac25)?

I also prefer `reputation-builder` to `reputation_builder` 👍

Issue #56, opened 2025-05-13
Implement Current Proposal Draft


carlaKC:
The latest full proposal for the mitigation we're looking at here is written up here: https://github.com/carlaKC/lightning-rfc/pull/5

Aiming to get this fully implemented in the simulator, so that we can start attacking it again. This is also a great chance to knock out any ambiguity in the spec document before we upstream it to the BOLTs PR.

A lot of it is already implemented, I believe the diff is:
* Congestion HTLCs should be endorsed #53 
* `effective_fees` should use the _outgoing_ endorsement signal for outgoing reputation (currently incorrectly impl-ed to use incoming signal)
* For outgoing reputation, general and congestion buckets should apply on the incoming link (currently impl-ed on the outgoing)*
* Abuse should be tracked on the outgoing channel (not incoming)
* Add slot restrictions to general bucket

The above roughly maps to the order I think we should do this, and each should be a a reasonably sized PR. It might be an incomplete list, so anything that comes up on read through should also be added!

* Likewise, incoming reputation should have a bucket on the outgoing link, but we're only dealing with outgoing reputation in this proposal.

elnosh:
notes so that I don't forget as I was going through the proposal

- currently the simulator allows access to congestion bucket if the htlc is endorsed and there are resources available but an htlc should only be granted access to congestion bucket if the `accountable` signal in the incoming htlc was **_not_** set. 
- if an htlc is unendorsed, we check if the peer has reputation and if it does, it allows access to `Protected` bucket. Before returning `Protected` and forwarding the htlc as `accountable`, we should check first if there are resources in the general bucket and if there is, forward without the `accountable` signal?

PR #65, opened 2025-05-28
use buckets on incoming channel

elnosh:
as part of https://github.com/carlaKC/jam-ln/issues/56, changed the bucketing to be applied on the incoming channel instead of outgoing. 

changes done:
- moved the bucket params from `OutgoingChannel` to `IncomingChannel`
- check misuse of congestion bucket on the outgoing channel instead of incoming
- added protected bucket to `IncomingChannel`
- slipped in a couple tests to `ForwardManager`

doing this before https://github.com/carlaKC/jam-ln/pull/62 so that I could add the protected bucket and use it for that PR. Also, the tests added in `ForwardManager` are incomplete but should be expanded on https://github.com/carlaKC/jam-ln/pull/62 when the correct bucketing outcomes are implemented.

### DIFF
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -501,8 +501,12 @@ impl ReputationManager for ForwardManager {
 mod tests {
     use std::time::{Duration, Instant};
 
-    use super::RevenueAverage;
-    use crate::ReputationParams;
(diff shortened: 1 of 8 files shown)


carlaKC (approved):
LGTM!

Since we're in a large refactor, going to prioritize getting these in for the sake of keeping up momentum. We can always come back and clean things up if we want (though they're looking pretty good rn!).

Issue #45, opened 2025-04-29
Implement "AddrMan" Style General Bucket


carlaKC:
As-is, general resources are trivially easy for an attacker to exhaust - all they have to do is forward as many non-dust HTLCs as we have slots and hold on to them, and they can jam up our resources for 2 weeks. This isn't the sort of behavior we expect from honest peers, so we implement a dos protection inspired by bitcoin core's address manager (that manages the IP addresses announced to it in the network).

---

For a general bucket with:
* `n` HTLC slots
* `m` liquidity assigned to the bucket

We divide our outgoing general bucket into `n` liquidity slots, where each slot holds `m/n` msat of liquidity.

Each of our incoming channels is assigned 5 liquidity slots that they may use (per outgoing channel):
* They can have at most 5 HTLCs in flight in the general bucket
* They can have at most `m*5/n` msat of liquidity in the general bucket

We'll assign these 5 slots as follows:
* Pick a random salt that will be used across all channels
* Choose 5 slots per channel `index = sha256(incoming short channel id + outgoing short channel id + salt + i) %n`

When a channel forwards us a HTLC that's bound for general, we'll check that resources are available only in the 5 indexes which represent their slots. If those slots are taken by other HTLC, we fail their HTLC back. If a HTLC is larger than `m/n`, it may take up as many liquidity slots as are available to that HTLC.

carlaKC:
Once this is done, it would be really interesting to run the simulator with no attack on the mainnet graph with payment activity turned right up and see whether any HTLCs end up being rejected due to these limits.

carlaKC:
Closed in #70

PR #70, opened 2025-06-06
Implement Resource Protection in General Resources

carlaKC:
Opening this up for discussion - not final design.

Possible approach for #45

### DIFF
--- a/ln-resource-mgr/src/lib.rs
+++ b/ln-resource-mgr/src/lib.rs
@@ -684,28 +684,13 @@ mod tests {
 
     #[test]
     fn test_forwarding_outcome_congestion() {
-        let mut check = test_congestion_check();
+        let check = test_congestion_check();
 
-        // Unaccountable htlc that is congestion elegible gets access to congestion bucket and is
+        // Unaccountable htlc that is congestion eligible gets access to congestion bucket and is
         // upgraded to accountable.
         assert!(
             check
                 .inner_forwarding_outcome(10, AccountableSignal::Unaccountable, true)
-                .unwrap()
-                == SuccessForwardOutcome {
-                    bucket: ResourceBucketType::Congestion,
-                    accountable_signal: AccountableSignal::Accountable
-                }
-        );
-
-        // Accountable htlc with sufficient reputation but no protected or general
-        // resources goes into congestion bucket and forwarded as accountable.
-        check.reputation_check.reputation = 1000;
-        check.resource_check.general_bucket.slots_available = 0;
-        check.resource_check.protected_bucket.slots_available = 0;
-        assert!(
-            check
-                .inner_forwarding_outcome(10, AccountableSignal::Accountable, true)
                 .unwrap()
                 == SuccessForwardOutcome {
                     bucket: ResourceBucketType::Congestion,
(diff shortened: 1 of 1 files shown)


ln-resource-mgr/src/forward_manager.rs
```diff
@@ -337,6 +340,7 @@ impl ReputationManager for ForwardManager {
                 v.insert(TrackedChannel {
                     capacity_msat,
                     incoming_direction: IncomingChannel::new(
+                        channel_id,
```
elnosh, line 343: wondering if we should also assign slots for every new added channel here in `add_channel`? iiuc we'll assign them at the time we are adding an htlc if the channel is not found in the `RestrictedBucket.candidate_slots` map.

ln-resource-mgr/src/lib.rs
```diff
@@ -278,7 +284,11 @@ impl AllocationCheck {
                     .general_bucket
                     .resources_available(htlc_amt_msat)
                 {
-                    Ok(ResourceBucketType::General)
+                    if self.general_eligible {
+                        Ok(ResourceBucketType::General)
+                    } else {
+                        Err(FailureReason::GeneralResourceLimit)
+                    }
```
elnosh, line 291: just to make sure this is the intended behavior, if the channel has exhausted its general slots, they don't get a shot at the congestion bucket?

ln-resource-mgr/src/incoming_channel.rs
```diff
@@ -20,19 +23,27 @@ pub(super) struct IncomingChannel {
     /// The resources available on the protected bucket. This will be used by htlcs that are
     /// accountable from peers that have sufficient reputation.
     pub(super) protected_bucket: BucketParameters,
+
+    pub(super) general_resources: RestrictedBucket,
```
elnosh, line 27: not sure what to suggest here but it feels weird to have a `general_bucket` and a `general_resources` field here in the incoming channel.

ln-resource-mgr/src/lib.rs
carlaKC: Discussed offline - we do indeed want to reject a pair that's taken up all of its slots even when general still has some space, this is how we get some DOS protections against people trying to spam up general.

ln-resource-mgr/src/forward_manager.rs
```diff
@@ -401,6 +405,7 @@ impl ReputationManager for ForwardManager {
 
         let fwd_outcome = match fwd_outcome {
             Ok(fwd_sucess) => {
+                // TODO: we need to add to the incoming channel here but it's not avail
```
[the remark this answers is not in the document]
carlaKC: > Also since we need to surface the AllocationCheck to report that htlc forward, I don't see a way around it bc we'd still need to do them separately (get the AllocationCheck to check the outcome and then add it)

Yeah indeed. Was wondering if we want to think of a different design that doesn't break things up into get check / then assign but I don't see an obvious way of avoiding it actually.

ln-resource-mgr/src/forward_manager.rs
carlaKC: > may_add_htlc won't need mutable references since they wouldn't be assigning the slots at the moment

Good point, will update 👍

ln-resource-mgr/src/lib.rs
```diff
@@ -263,6 +267,8 @@ impl AllocationCheck {
                             .general_bucket
                             .resources_available(htlc_amt_msat)
                         {
+                            // TODO: should we limit high reputation peers to their allocated
+                            // general resources?
```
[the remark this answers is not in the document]
carlaKC: Meaning, should we check `general_eligible` here. 

But on second look I don't think so - once somebody's got reputation we should basically just let them use whatever IMO.

elnosh:
all those allocations along with this which you had already mentioned before :man_facepalming: makes sense to go with opportunistic approach (:  

> it's unlikely that we need to use all of these allocations.

my bad I suggested you go down this path

carlaKC:
> all those allocations along with this which you had already mentioned before

Only gets real when you spend the time to realize it's big bad N-squared 🐺 

Always time well spent taking a look at alternatives, and we'll have a much better documented approach for having done it!

carlaKC:
Think I've addressed most of the review comments. Major changes since last review are:
* Stricter bucket assignments: eg, still require `general_eligible` if you have reputation because you may be the one attacking us
* Table driven test: added to try enumerate all `bucket_outcome` possibilities, left this in its own commit because it's big but can squash it if we want

ln-resource-mgr/src/lib.rs
```diff
@@ -764,73 +736,271 @@ mod tests {
...
+            {
+                let mut case = TestCase::new(
+                    "unaccountable_no_upgrade_congestion_available",
+                    AccountableSignal::Unaccountable,
+                    Ok(ResourceBucketType::Congestion),
+                );
+                case.upgradable = false;
+                case.set_congestion_available();
+                case
+            },
```
elnosh, line 929: not strictly related to this PR but looking at this test case it made me realize that this should instead be failing? Getting a bucket outcome of congestion will cause the HTLC to be upgraded to accountable but if it the htlc is not upgradable, it should fail.

elnosh (commented):
mostly LGTM

carlaKC:
* Added requirement for `upgradable` for congestion use: [diff](https://github.com/carlaKC/jam-ln/compare/10e69b2a8913d72773060e2d564487ddb1185aff..e289cba0c3b995c1bf6c7b12854a1687278f966f)
* Rebased to clear `Cargo.lock` conflict: [diff](https://github.com/carlaKC/jam-ln/compare/e289cba0c3b995c1bf6c7b12854a1687278f966f..951e7fd2f9e1ef708457d232ab2b10deb6197752)

elnosh (approved):

Issue #75, opened 2025-06-13
Bug: intercept_attacker_htlc will block attack by holding lock


carlaKC:
The `sink` attack implementation of `intercept_attack_htlc` [sleeps for a long time](https://github.com/carlaKC/jam-ln/blob/master/ln-simln-jamming/src/attacks/sink.rs#L142) to implement a slow jamming attack on a forwarding node.

We note on our attack trait that this function is allowed to block, because SimLN runs inerception in tasks. This is correct.
However, we've failed to account for our own blocking in `attack_interceptor.rs`:
* `intercept_htlc` locks our `attack` and calls [intercept_attack_htlc](https://github.com/carlaKC/jam-ln/blob/master/ln-simln-jamming/src/attack_interceptor.rs#L60)
* This lock will be held until we finish holding the HTLC
* Other interception calls will pile up until the HTLC is released

Ironically, this sort-of has the effect that we're going for anyway in this _particular_ attack (because it causes the attacker to hold HTLCs for long), but this isn't what we want in the more general case.

PR #76, opened 2025-06-16
ln-simln-jamming: remove mutex on attack interceptor

elnosh:
remove the `Mutex` on the attack interceptor. Should be up to each `JammingAttack` impl to be able to call `intercept_attacker_htlc` safely 

Fixes https://github.com/carlaKC/jam-ln/issues/75


carlaKC (approved):
✨

Issue #85, opened 2025-06-30
Bug: HTLC amount has been removed from bucket that doesn't contain it


carlaKC:
There's something going wrong with removing general bucket HTLCs. Full logs [here](https://drive.google.com/file/d/1_6RD_VcypAiM6344tl8lXVcfRziAIDwW/view?usp=sharing), too big to upload to GH. 

Can't immediately tell what the bug is, opening this up to start doing some digging (it's pretty consistently an issue). Interestingly on the few runs I've done, seems to often be happening on paths that have a hop with zero fees. 

So far as I can tell, this is the problematic payment:
```
2025-06-30T17:57:23.983Z DEBUG [simln_lib::sim_node] Got route: path 0:
 node_id: 0298f6074a454a1f5345cb2a7c6f9fce206cd0bf675d177cdbf0ca7508dd28852f, short_channel_id: 593736279064576, fee_msat: 0, cltv_expiry_delta: 144
 node_id: 033b63e4a9931dc151037acbce12f4f8968c86f5655cf102bbfa85a26bd4adc6d9, short_channel_id: 570646534881280, fee_msat: 0, cltv_expiry_delta: 34
 node_id: 02c4a7390c5c0e28eff5d8034a49d724900b196948f5723fecd887512b2f0020f3, short_channel_id: 433207581409280, fee_msat: 981, cltv_expiry_delta: 0
 blinded_tail: None
```

Forwarded by node 2:
```
2025-06-30T17:57:23.984Z INFO  [ln_simln_jamming::reputation_interceptor] Node 2 forwarding: 593736279064576:557 unaccountable -> 570646534881280 with fee 0 (981 -> 981) and cltv 144 (218 -> 74) with outcome forward as unaccountable
```

Then by node 13:
```
2025-06-30T17:57:24.130Z INFO  [ln_simln_jamming::reputation_interceptor] Node 13 forwarding: 570646534881280:1321 unaccountable -> 433207581409280 with fee 0 (981 -> 981) and cltv 34 (74 -> 40) with outcome forward as unaccountable
```

Then resolved:
```
2025-06-30T17:57:24.441Z INFO  [ln_simln_jamming::reputation_interceptor] Resolving htlc 570646534881280:1321 on 433207581409280 with outcome settled
2025-06-30T17:57:24.441Z INFO  [ln_simln_jamming::reputation_interceptor] Resolving htlc 593736279064576:557 on 570646534881280 with outcome settled
[...]

PR #87, opened 2025-07-08
fix duplicate slot remove bug

elnosh:
fixes https://github.com/carlaKC/jam-ln/issues/85

Tracking slots assigned by channel in `candidate_slots` to remove them by channel. Could use `HashMap` but left it as an array of tuples now since it's just 5 slots. 

Been running one and haven't hit `ErrBucketTooEmpty`

### DIFF
--- a/ln-resource-mgr/src/incoming_channel.rs
+++ b/ln-resource-mgr/src/incoming_channel.rs
@@ -65,11 +65,13 @@ pub(super) struct GeneralBucket {
     htlc_slots: Vec<bool>,
     /// Tracks the amount of liquidity allocated to each slot in the bucket.
     slot_size_msat: u64,
-    /// Maps short channel IDs to the slot indexes that a candidate channel is granted access to.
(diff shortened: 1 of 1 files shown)


ln-resource-mgr/src/incoming_channel.rs
```diff
@@ -437,4 +467,23 @@ mod tests {
 
         assert!(!bucket.add_htlc(scid, htlc_too_big).unwrap());
     }
+
+    #[test]
+    fn test_duplicate_remove_bug_addressed() {
```
carlaKC, line 472: _very_ personal preference: when fixing a bug, I like to add a test demonstrating that the bug is hit and then fix the bug + update the test in the next commit.

It creates a nice git history which clearly demonstrates how the bug was being hit, and makes the fixing of the bug a bit more test-driven. This isn't very common practice from what I've seen around, so def not blocking!

carlaKC (approved):
LGTM - would just like to add some asserts then this is gg!

Thanks for hunting and squashing this one 🔪

Issue #84, opened 2025-06-26
Allow running with/without attacker bootstrap


carlaKC:
There are two different types of setups that one may want when running an attack:
1) No attacker bootstrap: start with a network that has reputation for honest nodes, and the attacker's channels have no reputation
2) Attacker bootstrap: start with a network that has reputation for honest nodes, and and the attacker has built up reputation over some period

(right now we only allow the latter).

To make the simulator more easily runnable, we'll define the following standard structure for simualtions:
```
{network_name}
    simln_no_attacker.json // the graph without any attacker channels
    peacetime.csv // forwarding history for the graph without any attacker channels
    bootstrap_no_attacker.csv // summary of reputation for peacetime.csv
    simln_with_attacker.json // graph with attacker channels
    [optional] `bootstrap_with_attacker.csv // forwarding history for the graph with attacker channels
    [optional] `bootstrap_attacker_for_{t}.csv // summary of reputation in `bootstrap.csv` for period t (there can be multiple t values)
```

#### 1) Run without attacker bootstrap
To run the simulator:
`ln-simln-jamming --network-name X --target-alias Y --attacker-alias Z`

Simulator will look in `network-name` and use the following files:
- `sim_with_attacker.json`: the graph that we'll run our attack on
- `peacetime.csv`: for revenue projections
- `bootstrap_no_attacker.csv`: to bootstrap the reputation of the honest nodes, totally excluding the attacking node's channels.

Note: here we'll use the same files to bootstrap our network's reputation and project revenue. This duplication is _probably fine_, we can always add a requirement for another file if it becomes an issue.

#### 2) Run with attacker bootstrap
To run the simulator:
`ln-simln-jamming --network-name X --target-alias Y --attacker-alias Z --attacker-bootstrap 45d`

Simulator will look in `network-name` and use the following files:
- `sim_with_attacker.json`: the graph that we'll run our attack on
- `peacetime.csv`: for revenue projections
- `bootstrap_attacker_for_{t}.csv`: to bootstrap the reputation of the honest nodes, and the attacker for period `t`

Note: here the end user will need to generate their `bootstrap_{t}.csv` using `reputation-builder`, but they can generate multiple different periods from the same `bootstrap_with_attacker.csv`, which is nice.

carlaKC:
One thing to consider for the "no bootstrap" mode that I ran into in #89.

Right now, the simulator will shut down if the revenue during attack dips below the projected peacetime revenue:
[...]

Issue #72, opened 2025-06-10
Ideas: Make simulator more friendly to write attacks in


carlaKC:
Once we've got everything implemented, the goal will to be to hand this simulator on to other LN developers and ask them to try attack what we've got. Ideally, we should make this process as painless as possible, so that people actually spend some time trying to break this.

Opening this issue to make a list of possible actions, which we can spin out into issues if necessary.

carlaKC:
* Ability to call `send_to_route` on attacking nodes (depends on #52): we should provide access to `SimNode` so that people can implement active attacks
* Update readme to explain latest and greatest
* Check in some sample graphs + reputation files for quick start
* Utilities to generate own graph (#49 + graph building help)
* Ability to "prepay" reputation in attack: sometimes you want to start an attack assuming that you've already built reputation with a node by streaming payments through it (rather than passively sitting in the network w/ channels). It would be really cool for the attacker to be able to "prepay" reputation along a route when the simulation starts, and just be able to start with good reputation rather than needing to blast payment through (this just speeds things up - similar to the `general_jam_channel` utility that we have)
* Reporting on attacker costs (both reputation "prepay" and cost of sending payments)

Issue #71, opened 2025-06-10
Implement a No-Op Attack Strategy


carlaKC:
It will be useful to be able to run the simulator in the "steady state" where no attack is operating. This allows us to check things like:
- How [general bucketing](https://github.com/carlaKC/jam-ln/issues/45#issuecomment-2950386625) impacts the network
- See how reputation builds up over time
- Experiment with different bucket sizes / parameters

To do this, we can implement [Jamming Attack](https://github.com/carlaKC/jam-ln/blob/master/ln-simln-jamming/src/attacks/mod.rs#L23) on a `NoAttack` struct and allow it to run with the existing default implementation.

It would be great to be able to make a list of the different implemented attacks, and select them via CLI rather than having to re-build the simulation. Ie be able to run with `--attack {none/sink/ your attack name here}`.  This gets a bit ugly with async `dyn` traits, but would be nice to have if we can figure it out!

PR #88, opened 2025-07-10
jam attack impl

elnosh:
Hopefully a not so lame attempt at implementing an attack, tried doing one from the delving post. 

There are a couple things hardcoded and I added a couple extra channels to the base graph. 

Attack: jam and use all resources for a channel between [target_node_peer] <-> [target_node] by building reputation with attacker node. 

Using an attacker sender node (A1) and a attacker receiver (A2), will try to jam channel for:
A1 -> [target_peer] -> [target] -> A2 

- First use `build_reputation` helper to build reputation over A1 -> [target] -> A2 against the channel with the target peer.
- jam general resources.
- TODO: would need many channels to occupy slots in congestion (?)
- with sufficient reputation and general resources jammed, jam protected resources with continuous fast-failing payments that don't slash reputation. 

Need to define shutdown conditions. 

Other things for https://github.com/carlaKC/jam-ln/issues/72 to make it friendlier to write attacks:
- ability to define multiple attackers (and need to know how you will use them ?). I just hardcoded one of them.
- as mentioned here: https://github.com/carlaKC/jam-ln/issues/71 prob make an enum-like type to specify multiple attacks that the simulation could be run with. Although indeed ugly with `dyn` trait. Or could make a wrapper struct that impls `JammingAttack`. 
- there are some discrepancies with the default filenames used so they need manual modifying but I think they will be fixed in https://github.com/carlaKC/jam-ln/issues/84.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 target
 data
 results
-reputation-snapshots
+reputation-snapshots*
--- a/ln-simln-jamming/src/attacks/mod.rs
+++ b/ln-simln-jamming/src/attacks/mod.rs
@@ -9,6 +9,7 @@ use triggered::Listener;
 use crate::{accountable_from_records, records_from_signal, BoxError, NetworkReputation};
 
 pub mod sink;
+pub mod slow_jam;
 pub mod utils;
 
 pub struct NetworkSetup {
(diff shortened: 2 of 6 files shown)


elnosh:
Closing for https://github.com/carlaKC/jam-ln/pull/97

PR #89, opened 2025-07-14
Allow Running without Attacker Bootstrap

### LINKED ISSUE #84: Allow running with/without attacker bootstrap

carlaKC:
This PR introduces two key changes:
1. Network directory: enforce a fixed structure for the simulation files that we expect, so that end users can simply specify a `--network` option which will look for the files that we need.
2. No bootstrap: allow the simulation to run without any attacker bootstrap, only bootstrapping honest nodes and starting the attacker off with zero reputation.

Fixes #84

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *.json
 *.csv
+*.txt
 target
 data
-results
-reputation-snapshots
+results/*
--- a/README.md
+++ b/README.md
@@ -1,73 +1,102 @@
 # Jamming Simulator
 
-This simulator implements [outgoing reputation](https://gist.github.com/carlaKC/6762d88903d1cc27339859816ed80d43)
-for a network of nodes and performs a [sink attack](https://delvingbitcoin.org/t/hybrid-jamming-mitigation-results-and-updates/1147#p-3212-manipulation-sink-attack-9)
-against a target node in the network. In this attack, a malicious node
-will:
-- Open up channels and forward payments as usual to bootstrap its
-  reputation with a target node.
-- Slow jam the target's peers's general resources, so that all HTLCs
-  must be accountable to reach the target.
-- Hold any accountable HTLCs from the target with the goal of ruining its
-  reputation with its peers.
+This simulator implements the following to mitigate channel jamming
+attacks:
+- Outgoing reputation:
+  When forwarding a HTLC, the reputation of the *outgoing* channel is
+  compared to the revenue that's been earned on the *incoming* channel.
+- HTLC accountability: if a HTLC has occupied scarce resources, it
+  will be forwarded as accountable to indicate to the receiving node
+  that their reputation will be held accountable for its resolution.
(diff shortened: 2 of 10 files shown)


elnosh:
LGTM! Didn't run this for the 6months default tho. Only ran with `1d` duration.

elnosh:
I was able to start one with no attacker bootstrap. 

Just a note to put it somewhere, I ran the `forward-builder` for the 6 months default (for both attacktime and peacetime) but the csv files generated were around 150MB :thinking: Think these should've been 1GB+

carlaKC:
> Just a note to put it somewhere, I ran the forward-builder for the 6 months default (for both attacktime and peacetime) but the csv files generated were around 150MB 🤔 Think these should've been 1GB+

Also hit this, ~250 mb for 6 months.

Going to merge this as-is and investigate file size difference. Suspect that this may be due to general bucket restrictions being too strict.

PR #92, opened 2025-08-01
specify attack to run

elnosh:
Pulled this out from another branch implementing an attack. This adds a flag to the cli and a wrapper enum to be able to specify different attacks to run.  

It introduces a `AttackImpl` enum that acts as wrapper to any attack implementation. This makes it so that it implements the `JammingAttack` trait.

Any `JammingAttack` (Sink, SlowJam, NoAttack, etc) implementation can be added to the `AttackImpl` enum and run it for the simulation. 

This is a bit more code but no `dyn JammingAttack` and I think it makes it easy to just make your attack implementation and add it to the enum to run it.

It also slightly repurposed the `setup_for_network` method to `setup_for_attack` so that attacks that have no attacker bootstrap and are building reputation in it can run that before the simulation starts. So perhaps this method could run and only after call `simulation_completed` to avoid this https://github.com/carlaKC/jam-ln/issues/84#issuecomment-3084232538


elnosh:
Removed `AttackImpl` wrapper struct and changed to use `dyn JammingAttack` instead

ln-simln-jamming/src/attacks/mod.rs
```diff
@@ -27,10 +27,15 @@ pub struct NetworkSetup {
 #[async_trait]
 pub trait JammingAttack {
     /// Responsible for validating that the network provided meets any topological expectations for the attack, and
-    /// returning network-specific setup instructions for the attack.
+    /// returning network-specific setup instructions for the attack. Attack implementations can
+    /// also do any setup needed before starting the simulation such as building reputation for any
+    /// attacker nodes.
```
carlaKC, line 32: I don't think we should run things like reputation building in here, because we won't have started the simulation yet (there are no other payments in the network at this time).

`run_custom_actions` seems like a better home for that, since we spawn it in a task? Perhaps should be renamed `run_attack`.

ln-simln-jamming/src/main.rs
```diff
@@ -276,25 +255,45 @@ async fn main() -> Result<(), BoxError> {
...
-            .run_custom_actions(attacker_nodes, listener.clone())
+            .run_custom_actions(attacker_nodes, custom_actions_listener)
             .await
         {
             log::error!("Error running custom attacker actions: {e}");
             attacker_actions_shutdown.trigger();
         }
     });
 
+    // Spawn a task that will trigger shutdown of the simulation if the attacker loses reputation provided that the
```
carlaKC, line 281: While we're here update this comment? Now that we've moved the reputation check into the sink attack specifically.

ln-simln-jamming/src/parsing.rs
```diff
@@ -318,6 +323,52 @@ pub struct Cli {
...
+        AttackType::Sink => {
+            let attack = Arc::new(SinkAttack::new(
+                Arc::clone(&clock),
+                &sim_network,
+                simulation.target.1,
+                simulation.attacker.1,
+                risk_margin,
+                Arc::clone(&reputation_monitor),
+                Arc::clone(&revenue_monitor),
+                listener.clone(),
```
carlaKC, line 366: nit: unnecessary clones

carlaKC (commented):
Thanks for taking a look at `dyn` - think that this turns out really nicely with it! 

Just minor comments.

elnosh:
Resolved some conflicts

ln-simln-jamming/src/attacks/mod.rs
elnosh: > Polling a function is the way to do it for a sink attack, but for another type maybe you just want to run the whole attack and then call shutdown at the end, for example.

I think this approach still works fine for both types of attacks - either polling or if you want to run the attack and shutdown, you can do that and the shutdown would be sending on the channel. 

> just adding the instruction on `custom_actions` that you must call the shutdown trigger when you're done is the simplest way to go? + rename `custom_actions` to `run_attack`

I think both approaches (`custom_actions` + `simulation_completed` or `run_attack`) will work fine and yeah `run_attack` can be seen a bit simpler with instruction that implementation should shutdown itself whenever it's finished. Your call!

elnosh: renamed `run_custom_actions` to `run_attack` and removed `simulation_completed` check. So for sink attack `run_attack` mainly consists of doing the polling check that was previously done in `simulation_completed`

ln-simln-jamming/src/attack_interceptor.rs
```diff
@@ -116,8 +117,7 @@ mod tests {
             fn setup_for_network(&self) -> Result<crate::attacks::NetworkSetup, BoxError>;
             async fn intercept_attacker_htlc(&self, req: InterceptRequest) -> Result<Result<CustomRecords, ForwardingError>, BoxError>;
             async fn intercept_attacker_receive(&self,_req: InterceptRequest) -> Result<Result<CustomRecords, ForwardingError>, BoxError>;
-            async fn run_custom_actions(&self, attacker_nodes: HashMap<String, Arc<tokio::sync::Mutex<SimNode<SimGraph>>>>, simulation_completed_check: tokio::sync::oneshot::Sender<()>, shutdown_listener: Listener) -> Result<(), BoxError>;
-            async fn simulation_completed(&self, _start_reputation: NetworkReputation) -> Result<bool, BoxError>;
+            async fn run_attack(&self, _start_reputation: NetworkReputation, attacker_nodes: HashMap<String, Arc<tokio::sync::Mutex<SimNode<SimGraph>>>>, shutdown_listener: Listener) -> Result<(), BoxError>;
```
carlaKC, line 120: nitnit: if we'd done this refactor before the core change, would be able to have slightly less churn here (wouldn't need to add and remove the oneshot).

More a general structural comment than something that needs to change here!

ln-simln-jamming/src/attacks/sink.rs
```diff
@@ -250,44 +247,59 @@ where
...
+                            snapshot.peacetime_revenue_msat,
+                            snapshot.simulation_revenue_msat,
+                            snapshot.runtime
+                        );
+
+                        return Ok(());
+                    }
 
-            return Ok(true);
+                    log::trace!(
```
carlaKC, line 274: We don't actually have cli-configurable logging, so there isn't a way to run the simulation that would log this.

Happy to change it to debug or surface logging options in CLI (can be a follow up).

ln-simln-jamming/src/main.rs
```diff
@@ -267,58 +266,26 @@ async fn main() -> Result<(), BoxError> {
...
         .collect();
 
-    let attacker_actions_shutdown = shutdown.clone();
-    let attack_custom_actions = Arc::clone(&attack);
-    let custom_actions_listener = listener.clone();
-    // Channel to signal when the attack is ready to start calling the `simulation_completed`
-    // check. Sender will be passed to `run_custom_actions` and signal from there that the check
-    // for the attack shutdown criteria can be started.
-    let (tx, mut rx) = oneshot::channel();
+    let main_attack = Arc::clone(&attack);
```
carlaKC, line 269: Unnecessary clone, we can allow `attack` to be moved by the task spawn because it's not needed again.

carlaKC (approved):
Two tiny nits, otherwise LGTM!

ln-simln-jamming/src/attacks/sink.rs
elnosh: Think I'd rather add the logging option in CLI for other log levels. Will do in follow-up :thumbsup:

PR #94, opened 2025-08-05
Handle shutdown and write target's forwards to results dir

elnosh:
If terminating simulation before reaching any of the attack-specific shutdown conditions (ctrl-c) handle the shutdown and write any pending records from the `results_writer`.

Also, think it'd be useful to write the target's forward records in addition to the ones already being written.

### DIFF
--- a/ln-simln-jamming/src/parsing.rs
+++ b/ln-simln-jamming/src/parsing.rs
@@ -140,14 +140,15 @@ impl SimulationFiles {
         // We only allow one target node, but if there are multiple aliases in this file we'll
         // fail to find the pubkey by alias below.
         let target_alias = fs::read_to_string(network_dir.join("target.txt"))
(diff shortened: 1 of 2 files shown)


ln-simln-jamming/src/parsing.rs
```diff
@@ -141,13 +141,13 @@ impl SimulationFiles {
         // fail to find the pubkey by alias below.
         let target_alias = fs::read_to_string(network_dir.join("target.txt"))
             .map_err(|e| format!("target.txt file containing target alias not found: {}", e))?;
+        let target_alias = target_alias.trim();
```
carlaKC, line 144: nit: just inline the `trim` with the line above?

carlaKC (approved):

ln-simln-jamming/src/parsing.rs
elnosh: done

PR #95, opened 2025-08-06
Trigger simulation shutdown

elnosh:
Sorry I missed this from previous PR.

If `simulation_completed` for the attack returns true then also trigger the simulation to shutdown. I noticed the simulation was not stopping even if the shutdown criteria was met.


carlaKC (approved):
LGTM, sorry I missed this on review.

We really should be able to pass the `shutdown` + `listener`  into `create_simulation_with_network` so that we can share one signal, but that needs a sim-ln change, so happy to merge this as-is.

PR #96, opened 2025-08-07
Some helpers

elnosh:
Some small things:
- helper to jam congestion resources
- Write bucket results to the reported forwards (thought this is useful)
- Changed `build_reputation` helper to calculate the htlc risk for the number of htlcs we want to build reputation for instead of just 1.


ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -135,7 +135,8 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
 
         htlc_risk = params
             .reputation_params
-            .htlc_opportunity_cost(target_hop.fee_msat, target_hop.cltv_expiry_delta);
+            .htlc_opportunity_cost(target_hop.fee_msat, target_hop.cltv_expiry_delta)
+            * params.htlcs.len() as u64;
```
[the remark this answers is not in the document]
elnosh: I made the stupid assumption that the amounts in `htlcs` will all be the same... 
Will fix and add the tests :thumbsup:

elnosh:
pushed changes addressing comments

carlaKC (commented):
Still some questions about `reputation_builder` - I think a more robust test will help us make sure it's doing what we want (esp when we get too brain-tangled to even know what this is ourselves).

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -186,3 +187,132 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+        }
+    }
+
+    let threshold = if channel_revenue - peer_reputation > 0 {
+        channel_revenue - peer_reputation
+    } else {
+        0
+    } as u64;
+
+    let fee_to_bump_reputation = threshold + htlc_amounts + htlc_risk + fee_buffer;
```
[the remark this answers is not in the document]
elnosh: no valid reason, just a flaw in my understanding. Hopefully should be correct now with test added

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+            if let Some(target_idx) = path
+                .hops
+                .iter()
+                .position(|hop| hop.pubkey == target_hop_pubkey)
+            {
+                // Add up the cltv_delta(s) from the target hop to get the total number of blocks
+                // until the htlc can be resolved. Here we assume the current height is 0.
+                //
+                // We use `build_route_from_hops` from LDK to build the routes for the
+                // payments. However, that method adds a random CLTV offset so the route we got
```
carlaKC, line 205: Wow, surprised that we can't turn this off :|

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+                // here could have a different `cltv_expiry_delta` from a route we got another time
+                // after we have built reputation. Differences in the cltv deltas in some cases
+                // could case the calculated fee here to not be enough. Hence, we default to
+                // a reasonably higher value of 200 which overpays a little but should be enough
+                // to cover the random offsets.
+                let cltv_expiry: u32 = path.hops[target_idx..].iter().len() as u32 * 200;
```
carlaKC, line 211: If the 200 is to account for a "shadow" hop, shouldn't this be + 200 rather than multiply?

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+                total_htlc_risk += current_htlc_risk;
+            }
+        }
+    }
+
+    let threshold = if channel_revenue - peer_reputation > 0 {
+        channel_revenue - peer_reputation
+    } else {
+        0
+    } as u64;
```
carlaKC, line 226: Shouldn't this include the `total_htlc_risk`

Eg:
* `channel_revenue` = 100
* `peer_reputation` = 150
* `total_htlc_risk` = 40

`threshold` = 0 (since 100 -150 < 0), but then we'll go on to pay 40 for `total_htlc_risk` even though we actually have enough reputation for it?

Also _could_ technically overflow to use `saturating_sub`?

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+            reputation_params: ReputationParams {
+                revenue_window: Duration::from_secs(60),
+                reputation_multiplier: 60,
+                resolution_period: Duration::from_secs(90),
+                expected_block_speed: None,
+            },
+            general_slot_portion: 30,
+            general_liquidity_portion: 30,
+            congestion_slot_portion: 20,
+            congestion_liquidity_portion: 20,
```
carlaKC, line 318: Could use `default()` here if you like

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+                    fwd_params.htlc_opportunity_cost((*htlc as f64 * fee_pct) as u64, expiry_delta);
+            }
+            risk
+        };
+
+        // The fee that we expect to pay is the sum of:
+        // - Revenue/reputation diff -> 1000000 - 700000 = 300000
+        // - Sum of risk of each htlc
+        // - Fee buffer -> 1000
+        let expected_fee = 300_000 + expected_htlc_risk + 1000;
```
carlaKC, line 377: Perhaps add tests cases for:
- We have enough reputation and don't need to pay
- We have reputation that covers part of the `expected_htlc_risk` but not all 

Should be easy enough to turn this into a multi-case test.

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -176,9 +172,371 @@ pub async fn build_reputation<C: Clock + InstantClock, R: ReputationMonitor>(
...
+
+        assert_eq!(fee_to_pay, expected_fee);
+        assert_eq!(total_htlc_risk, expected_htlc_risk);
+    }
+
+    #[tokio::test]
+    async fn test_build_reputation() {
```
carlaKC, line 384: Great test 🔥

carlaKC (commented):
Sorry this took me so long!

ln-simln-jamming/src/attacks/utils.rs
elnosh: gave a glance at the ldk method that adds the offset and saw this comment `this method obfuscates the CLTV expiry deltas along the payment path by adding a randomized 'shadow route' offset to the final hop.`

And here https://github.com/lightningdevkit/rust-lightning/blob/96f9242465d660a58b38302b3e7e945fea4465e4/lightning/src/routing/router.rs#L3807-L3815 is where it looks like it does that.

So yeah I think we can also add up all the real cltv expiry deltas and add an extra high offset for this shadow hop

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -370,14 +345,65 @@ mod tests {
             risk
         };
 
-        // The fee that we expect to pay is the sum of:
-        // - Revenue/reputation diff -> 1000000 - 700000 = 300000
-        // - Sum of risk of each htlc
-        // - Fee buffer -> 1000
-        let expected_fee = 300_000 + expected_htlc_risk + 1000;
+        let cases = vec![
```
carlaKC, line 355: Nice, thanks for adding these!

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -336,11 +339,15 @@ mod tests {
         let fee_pct = 0.0001;
 
         let htlc_amounts = vec![21_000, 42_000, 100_000];
+
```
carlaKC, line 342: nit: unnecessary newline, remove on squash

ln-simln-jamming/src/attacks/utils.rs
```diff
@@ -336,11 +339,15 @@ mod tests {
...
         let htlc_risk = {
             let mut risk = 0;
             for htlc in htlc_amounts.iter() {
-                risk +=
-                    fwd_params.htlc_opportunity_cost((*htlc as f64 * fee_pct) as u64, expiry_delta);
+                // Test routes we are using this test are just one hop but multiply by 2 to account
+                // for the extra offset added.
+                risk += fwd_params
+                    .htlc_opportunity_cost((*htlc as f64 * fee_pct) as u64, expiry_delta)
+                    * 2;
```
carlaKC, line 350: Feel like we've picked a magic number in this test that makes this line correct?
* We have a delta of 200
* In the code we add 200
* Here we multiply by 200 and it _happens_ to be equivalent to adding 200

Make `expiry_delta` a less magical number (80) and pull the offset we add into a constant?

carlaKC (approved):
lgtm - just one nit on the test values, can be fixed on squash!

ln-simln-jamming/src/attacks/utils.rs
elnosh: Indeed, it was kinda confusing and only worked because I set the `expiry_delta` to 200. Changed the value and pulled the 200 offset into a const.

PR #97, opened 2025-08-07
Fast jam attack impl 

elnosh:
Attack: jam and use all resources for a channel between [target_node_peer] <-> [target_node] by building reputation with attacker node.

Using an attacker sender node (A1) and a attacker receiver (A2), will try to jam channel for:
A1 -> [target_peer] -> [target] -> A2

- First use build_reputation helper to build reputation over A1 -> [target] -> A2 against the channel with the target peer.
- Jam general resources and congestion resources with helpers. 
- With sufficient reputation and general resources jammed, jam protected resources with continuous fast-failing payments that don't slash reputation.
       - Depending on the projected simulated forwards and the reputation snapshot, most likely the [target] does not have enough reputation with [target_peer] to take up protected slots, it can only take the 5 general slots that [target_peer] would allow [A1]. So to take all protected resources it would need many different incoming channels. 

<s>It builds upon https://github.com/carlaKC/jam-ln/pull/92 and https://github.com/carlaKC/jam-ln/pull/96 so you can look at the last commit only https://github.com/carlaKC/jam-ln/commit/ea15d89d312d94dde7944fd3dafb9e25c432873c for the actual attack impl. </s>

I added a couple extra channels to a graph I had.

### DIFF
--- a/ln-simln-jamming/src/attacks/fast_jam.rs
+++ b/ln-simln-jamming/src/attacks/fast_jam.rs
@@ -1,0 +1,344 @@
+use crate::{
+    attacks::JammingAttack,
+    clock::InstantClock,
+    print_request,
+    reputation_interceptor::{ChannelJammer, ReputationMonitor},
+    revenue_interceptor::PeacetimeRevenueMonitor,
+    BoxError, NetworkReputation,
+};
+
+use async_trait::async_trait;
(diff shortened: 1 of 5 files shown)


elnosh:
building reputation is a bit clunky still. With a clock speedup of 200 it seems to work but above that it fails.

PR #98, opened 2025-08-08
Pick live timestamp after main lock is acquired to prevent "drift" during speedup

### LINKED ISSUE #80: Bug: Very fast speedup causes last updated issue
When running data generation with #79 with a very high speedup (1000) I ran into the following error:
```
2025-06-24T18:05:34.300Z ERROR [simln_lib::sim_node] Could not remove htlcs from channel: InterceptorError: last updated reputation at Instant { tv_sec: 1670248, tv_nsec: 612543916 }, read at Instant { tv_sec: 1670203, tv_nsec: 583418916 }.
```

I can somewhat consistently hit this with forward generation, but I think that I've also seen it in simulations with high speedup:
* Graph file: [simln_bootstrap.json](https://github.com/user-attachments/files/20889226/simln_bootstrap.json)
* `forward-builder --sim-file data/lvl_acinq/simln_bootstrap.json --duration=1w --target-alias 24 --attacker-alias 70`

This matters for data generation, because generally we want about 6 months of data. The difference between a 500 (works fine) and 1000 (max) speedup is significant - 4 hours vs 8 :|

carlaKC:
Fixes #80, should make data generation a whole lot less painful!

### DIFF
--- a/ln-simln-jamming/src/bin/forward_builder.rs
+++ b/ln-simln-jamming/src/bin/forward_builder.rs
@@ -59,7 +59,7 @@ async fn main() -> Result<(), BoxError> {
 
     let network_dir = SimulationFiles::new(cli.network.network_dir.clone(), cli.traffic_type)?;
 
-    let clock = Arc::new(SimulationClock::new(500)?);
+    let clock = Arc::new(SimulationClock::new(1000)?);
     let tasks = TaskTracker::new();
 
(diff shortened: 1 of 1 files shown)


elnosh (approved):
code LGTM. Ran one for 6 months with the 1000 clock speedup and didn't hit the `ERROR [simln_lib::sim_node] Could not remove htlcs from channel: InterceptorError: last updated reputation at Instant` error.

PR #100, opened 2025-08-14
ln-simln-jamming: update sim-ln dependency

carlaKC:
Update to include two important changes:
- Determinism fix: previously lock acquisition ordering could impact the order that payments are sent, this is fully fixed now for a fixed seed run of the simulator.
- Pathfinding feedback: sim-node now reports success/failures to its pathfinder, which makes for more realistic behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2500,7 +2500,7 @@ dependencies = [
 [[package]]
 name = "sim-cli"
 version = "0.1.1"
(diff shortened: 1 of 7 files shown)


carlaKC:
@elnosh do you think it's worth making everything generic over `Clock + InstantClock` vs just hardcoding the `SimualtionClock` type? 

Originally made this generic because I thought we might want to mock out the clock, but hasn't really turned out that way.

carlaKC:
[Rebased](https://github.com/carlaKC/jam-ln/compare/64c34a79a058d254e1b4998b6a7d3eb616f44f23..084085b394902cdbfb1db9f28fb1dcd7a6214a81)

elnosh:
lgtm

PR #106, opened 2025-09-16
Scale htlc risk by CLTV delta

carlaKC:
We're struggling to get reputation because multiplying htlc fees by the total time it could be held by (cltv delta, assuming 10 minute blocks, expressed in seconds) is a very large number.

To capture the idea of a "reasonable" hold time, we divide hold time by our node's chosen CLTV delta. This allows us to increase risk as hold time increases, without having to have such large numbers.

The unfortunate implementation detail here is that we now have to have our CLTV delta handy, which is a bit of a hack. Opening early in draft, last commit could probably use some work.

### DIFF
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -135,14 +135,6 @@ impl Default for ForwardManagerParams {
     }
 }
 
-impl ForwardManagerParams {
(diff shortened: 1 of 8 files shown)


carlaKC:
Closing this for now as we're good with #105 at the moment.

If we find that we can't build reputation IRL, this would be a good approach to scale this value down a bit.

PR #109, opened 2025-09-16
ln-resource-mgr: bump number of assigned slots in general channel

carlaKC:
The 5 slot limit was calculated for a V3 world - bump this limit up because we're still living in a world with 483 slots.


carlaKC:
cc @elnosh, merge it if you need it!

elnosh (approved):

PR #114, opened 2025-09-18
slow jam attack impl

elnosh:


ln-simln-jamming/src/attacks/slow_jam.rs
```diff
@@ -0,0 +1,329 @@
...
+
+#[async_trait]
+impl<R, J> JammingAttack for SlowJam<R, J>
+where
+    R: ReputationMonitor + Send + Sync,
+    J: ChannelJammer + Send + Sync,
+{
+    fn setup_for_network(&self) -> Result<NetworkSetup, BoxError> {
+        // Validate that attacker receiver has channel with target.
+        self.target_channels
```
carlaKC, line 191: I think that we should add some stricter checks here, just so that we can be sure that everything is happening as expected:
* `channel_to_jam` is in `target_channels`
* `attacker_sender` has a channel with the node that owns `channel_to_jam`
* The `sanity_check_node` has a channel with `channel_to_jam` node
* The honest `receiver_pubkey` has a channel with `target`

Specifically the last one, because this is how we check we've succeeded with a test payment. If we don't have a `target -- receiver_pubkey` channel, we'll still get a failed payment and think that we succeeded but actually it was just a bad route.

ln-simln-jamming/src/attacks/slow_jam.rs
```diff
@@ -0,0 +1,329 @@
...
+        let payment_hash = PaymentHash(rand::random());
+        if let Err(e) = attacker_node_sender
+            .lock()
+            .await
+            .send_to_route(route, payment_hash, None)
+            .await
+        {
+            return Err(e.to_string().into());
+        }
+        self.jamming_payments.lock().await.insert(payment_hash);
```
carlaKC, line 177: The timing of this seems unlikely, but I think it's _possible_ that we get a race where:
* We dispatch the payment
* `intercept_attacker_receive` gets the `jamming_payments` lock
* Payment is not held because not yet added
* We get the `jamming_payments` lock and add it

Can be addressed by just moving this addition up to above the payment dispatch (and can be deleted if we fail to send).

ln-simln-jamming/src/attacks/slow_jam.rs
```diff
@@ -0,0 +1,329 @@
...
+        // After building reputation and jamming general resources, jam protected resources.
+        self.slow_jam_channel(&attacker_nodes).await?;
+
+        // Receiver of an honest payment
+        let receiver_pubkey = PublicKey::from_str(
+            "03ae641aa89c4d2c0034b31bad23a4f80f841d593ddd52f7f35cf7896d0ef8a594",
+        )?;
+        // With protected resources jammed, try making a test payment and check that it fails.
+        let test_hops = vec![self.channel_to_jam.0, self.target_pubkey, receiver_pubkey];
+        let sanity_check_route = build_custom_route(
```
carlaKC, line 275: Little odd that the `receiver_pubkey` is hardcoded but we pass the `sanity_check_node` in? Let's make both parameters and then have all our hard coding in `parsing.rs`?

ln-simln-jamming/src/parsing.rs
```diff
@@ -352,14 +370,79 @@ where
...
+    Ok(Arc::new(
+        populate_network_graph(channels, clock.clone())
+            .map_err(|e| SimulationError::SimulatedNetworkError(format!("{:?}", e)))?,
+    ))
+}
+
 impl Cli {
     pub fn validate(&self) -> Result<ForwardManagerParams, BoxError> {
-        if self.target_reputation_percent == 0 || self.target_reputation_percent > 100 {
+        //if self.target_reputation_percent == 0 || self.target_reputation_percent > 100 {
```
carlaKC, line 444: nit: delete commented out

carlaKC (commented):
Looking good! I think we just need to tighten up validation a bit and then this will be good to go.

In a follow up, I think that it would be interesting to re-use this logic to do slot jamming (rather than the liquidity jamming that we have here) - it'll probably end up cheaper (reputation for ~200 above-dust htlcs should be cheaper than the full liquidity of the channel).

ln-simln-jamming/src/attacks/slow_jam.rs
elnosh: indeed. I ended up hardcoding many variables that it got mixed up. Will move to `parsing.rs`

ln-simln-jamming/src/attacks/slow_jam.rs
elnosh: I completely blanked out on fact that I had the `network_graph` :man_facepalming:
Added checks for these two: 

> - attacker_sender has a channel with the node that owns channel_to_jam
> - The sanity_check_node has a channel with channel_to_jam node

carlaKC (approved):
LGTM!

Just one nit, feel free to ignore or add on squash. Feel free to go ahead with merge once squashed 🥇

PR #115, opened 2025-09-18
Add option to run with no attack and report network stats

carlaKC:
It's useful to be able to run the simulator without any attack, do do things like see what the impact of general bucketing is on the network at different levels of activity/expected payment size. 

While we're here, add reporting on the forwarding outcomes of each htlc in the network, so it's easy to get an idea of network-wide differences.


carlaKC:
Some of the refactoring in here is still nice, but we don't need the no-attack option (simpler to just run in forward-builder).

PR #117, opened 2025-09-22
Add ability to run with/without reputation in forward-builder

carlaKC:
It's interesting to be able to run the network with reputation enabled/disabled to get an idea of the impact of the reputation system on the network in times of peace. 

Rather than add this to our main simulation (which has an attack), this PR adds an option to the forward-builder tool to just be able to run the simulation.

Right now, we just run the simulation with no additional analysis, which provides us with payment count / success rates. If we're interested in more information we can possibly expand the interceptors that we use to track more information.


carlaKC:
Don't need right now.

PR #120, opened 2025-09-30
increase opportunity cost gradually

### LINKED ISSUE #119: Switch from stepwise to gradual oppotrunity cost
We have updated the spec to deal with an edge case in how we deal with opportunity cost:
- Previously: `opportunity_cost = ceil ( (resolution_time - resolution_period) / resolution_period) * fees`
- Now: `opportunity_cost = max(0,  (resolution_time - resolution_period) / resolution_period) * fees`

We should update our simulation accordingly.

elnosh:
Closes https://github.com/carlaKC/jam-ln/issues/119

### DIFF
--- a/ln-resource-mgr/src/htlc_manager.rs
+++ b/ln-resource-mgr/src/htlc_manager.rs
@@ -32,7 +32,11 @@ impl ReputationParams {
     /// Calculates the opportunity_cost of a htlc being held on our channel - allowing one [`reputation_period`]'s
     /// grace period, then charging for every subsequent period.
     pub(super) fn opportunity_cost(&self, fee_msat: u64, hold_time: Duration) -> u64 {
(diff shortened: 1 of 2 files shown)


carlaKC (approved):
LGTM!

Going to hold off on merging this for now, as it'll change any results we get from simulation runs. Want to get those out the door and then we can hit the button.

GeorgeTsagk:
> Going to hold off on merging this for now, as it'll change any results we get from simulation runs. Want to get those out the door and then we can hit the button.

Could merge it in combination with #130. Will allow progression with new tweaks in our experiments while keeping the defaults the same for any newcomers.

carlaKC:
cc @elnosh would you please merge #120 and do a review pass on this while I'm out 🙏

elnosh:
think it's ready as-is, merging

PR #129, opened 2025-10-16
fix sleep when processing peacetime forwards

elnosh:
still running simulation but I think this should fix it to sleep for correct duration

### DIFF
--- a/ln-simln-jamming/src/revenue_interceptor.rs
+++ b/ln-simln-jamming/src/revenue_interceptor.rs
@@ -163,7 +163,14 @@ impl RevenueInterceptor {
     /// Replays projected forwards for the simulated network in a time of peace (ie, without the attacker). These events
     /// are replayed "live" so that they can be compared to the current simulation's revenue.
     pub async fn process_peacetime_fwds(&self) -> Result<(), BoxError> {
-        let mut last_event_ts = None;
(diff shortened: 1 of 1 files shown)


ln-simln-jamming/src/revenue_interceptor.rs
```diff
@@ -163,7 +163,14 @@ impl RevenueInterceptor {
     /// Replays projected forwards for the simulated network in a time of peace (ie, without the attacker). These events
     /// are replayed "live" so that they can be compared to the current simulation's revenue.
     pub async fn process_peacetime_fwds(&self) -> Result<(), BoxError> {
-        let mut last_event_ts = None;
+        let mut last_event_ts = self
+            .peacetime_revenue
+            .lock()
+            .await
+            .revenue_events
+            .pop()
```
carlaKC: nit: use `peek` here instead? we're always going to drop our first event otherwise

carlaKC (approved):
nice catch!

ln-simln-jamming/src/revenue_interceptor.rs
elnosh: oops yes, fixed.

carlaKC (approved):

PR #130, opened 2026-06-03
Pluggable reputation algorithms + experiment labels

GeorgeTsagk:
## Description

Makes it cheap to add and compare reputation scoring algorithms in the simulator.

  A `ReputationAlgo` enum on `ReputationParams` selects the opportunity-cost scoring rule, dispatched in `opportunity_cost`. Adding an algorithm is now a localised change: one enum variant, one match arm, one CLI value. Two are provided:

  - **`original`** (default) — the pre-existing stepwise cost `floor(hold/period) × fee`.
  - **`gradual`** — a continuous cost `max(0, (hold − period)/period) × fee`: linear above the resolution period, zero below, no stair-step at each boundary (same formula as #120, absorbed that commit on this PR to demonstrate the addition).

Select via `--reputation-algo` (default `original`). A new `--label` flag namespaces output to `results/{Attack}/{label}/{timestamp}` (label defaults to the algo name), so runs of the same experiment group together. The only change to existing runs is the results path layout — scoring is untouched at the default.

### Rationale

Want to experiment with a few adjustments to the HTLC opportunity cost formula, which could (possibly) help us mitigate fast jamming too. This PR was created in order to make comparison of different algos a more easy and straight forward process.

### DIFF
--- a/ln-resource-mgr/src/forward_manager.rs
+++ b/ln-resource-mgr/src/forward_manager.rs
@@ -3,7 +3,7 @@ use crate::incoming_channel::{BucketParameters, IncomingChannel};
 use crate::outgoing_channel::OutgoingChannel;
 use crate::{
     AllocationCheck, BucketResources, ChannelSnapshot, ForwardResolution, ForwardingOutcome,
-    HtlcRef, ProposedForward, ReputationCheck, ReputationError, ReputationManager,
+    HtlcRef, ProposedForward, ReputationAlgo, ReputationCheck, ReputationError, ReputationManager,
     ReputationParams, ResourceBucketType, ResourceCheck,
 };
 use std::collections::hash_map::Entry;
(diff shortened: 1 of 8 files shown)


elnosh (commented):
seems fine to me if you want test different opportunity cost function. I just merged https://github.com/carlaKC/jam-ln/pull/120 so can do the gradual as default and then add any other you are experimenting with.

a bit too much llm-verbosity here 😅 
```
--reputation-algo <REPUTATION_ALGO>
     The reputation scoring algorithm to use. Defaults to the pre-existing `original`

          [default: original]

          Possible values:
          - original: Pre-existing stepwise opportunity cost
          - gradual:  Gradual opportunity cost (bolts#1280 / jam-ln#119): linear above the resolution period, zero below, no stair-steps
```

GeorgeTsagk:
Closing in favor of https://github.com/carlaKC/jam-ln/pull/132

PR #132, opened 2026-06-12
Add new attack types (`slow-slot`, `fast-slot`, `fast-liq`)

GeorgeTsagk:
## New jamming profiles and a reputation fix for sub‑period holds (<=90s)

  Adds three protected‑bucket jamming attacks, a pluggable opportunity‑cost framework, and a new algorithm that attempts to close the gap they exploit.

  Attacks (`--attack-type`):
  - `slow‑slot`: fill the protected slot bucket with dust HTLCs, hold for weeks.
  - `fast‑slot`: same fill, but hold ~85s, cancel, repeat.
  - `fast‑liq`: exhaust protected liquidity with a few large HTLCs, ~85s, repeat.

  Why they work: the gradual opportunity cost `max(0, (hold − period)/period)·fee` is 0 for any hold below the 90s resolution period. So an attacker holding at 85s cancels and re‑jams indefinitely with zero reputation penalty: the jam is essentially free to sustain. 

> A 3‑month slow‑slot run denies ~98% of a channel's revenue at ~0.5× the attacker's cost, which is almost entirely the one‑time entry.

Run it locally: one command, generates traffic + reputation from scratch:
```
  make slow-slot-jam        # also: fast-slot-jam, fast-liq-jam
```

  The new algo `Ramp`: a hold ≤90s should not be free. Ramp charges nothing below a 5s grace, ramps linearly to one full fee at the period, then continues at the gradual rate: so sustained sub‑period jamming bleeds reputation and costs.

Selectable via `--reputation-algo {original,gradual,ramp}`; default behaviour unchanged.

### Notes

- This PR is a placeholder for experimenting: tweaking and re-running experiments takes multiple hours. When demonstrating attacker cost vs denied revenue sometimes you have to go beyond 2w as break-even happens at a later point. (~7-8h of sims).

- I will keep this as draft, then share numbers for all attacks vs original algo & all attacks vs `Ramp` algo (mitigates fast variants).

- Instead of creating a lot of small PRs that are hard to argue about, I'll just connect everything here, which is going to be easier to monitor.

- I removed the base fee for this simulation, as that alone makes the slot attacks very expensive. Reputation based defense should work well with 0 base fees, which is a network trend as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,11 @@ data
 results/*
 reputation_*
 revenue_*
+
+# Generated simulation data (rebuilt on demand by the Makefile / setup script).
+networks/*/peacetime_traffic.csv
+networks/*/reputation.csv
+# Attack-time networks the setup script derives from the tracked SlowJam template.
+networks/*/attacks/SlowSlotJam/
+networks/*/attacks/FastSlotJam/
+networks/*/attacks/FastLiqJam/
--- a/Makefile
+++ b/Makefile
@@ -18,3 +18,101 @@ install-tools:
 
 install:
 	cargo install --locked --path ln-simln-jamming
+
+# ============================================================================
+# Slot / liquidity jamming experiments — one command, data grabbed automatically
+# ============================================================================
+# `make <attack>` builds the binaries, generates the peacetime traffic + reputation
+# snapshot if they're missing (cached afterwards), positions the attacker, runs the
+# attack, and prints the summary. Override any VAR=value:
+#
+#   make slow-slot-jam                                   # quick demo (~20 min first run)
+#   make slow-slot-jam JAM_DURATION=8w TRAFFIC_DURATION=10w   # the headline economics
+#   make fast-slot-jam REPUTATION_ALGO=gradual           # compare an algorithm
+#   make sim-clean-data                                  # drop generated data to regenerate
+#   make sim-help                                        # list the knobs
+
+NET               ?= networks/ln_slow_jam
+SPEEDUP           ?= 200
+# How long to sustain the jam. TRAFFIC_DURATION must be >= JAM_DURATION (the live revenue
+# counterfactual replays the raw traffic file over the jam).
+JAM_DURATION      ?= 1d
+# Peacetime traffic to generate. Kept short: reputation-builder --allow-boost tiles it to fill the
+# (unchanged, protocol-default) 6-month reputation window, so a week of traffic bootstraps full
+# reputation without generating months of data. Grow it for longer jams.
+TRAFFIC_DURATION  ?= 1w
+REPUTATION_ALGO   ?= original
+# Experiment placement (defaults reproduce the busy, low-base-fee target in ln_slow_jam):
+TARGET_ALIAS      ?= 2
+CHANNEL_TO_JAM    ?= 570646534881280
+SENDER_ALIAS      ?= 25
+RECEIVER_ALIAS    ?= 70
+# Fee on the target->receiver channel (the hop the held HTLCs exit through) — drives the attacker's
+# one-time entry cost via htlc_risk, not the denied revenue. We model 0-base-fee target channels (a
+# common LN policy): with 0 base, dust slot-jam HTLCs carry ~no fee so their htlc_risk ~ 0 and entry
+# collapses to the revenue threshold. The proportional fee still gates large liquidity-jam HTLCs.
+RECEIVER_BASE_FEE ?= 0
+RECEIVER_PROP_FEE ?= 100
+LABEL             ?= $(ATTACK)
+
+SIM   := ./target/release/ln-simln-jamming
+FWD   := ./target/release/forward-builder
+REP   := ./target/release/reputation-builder
+SETUP := python3 ln-simln-jamming/scripts/setup_attack_network.py
+
+.PHONY: bins
+bins:
+	cargo build --release --bins
+
+# Prerequisites generated on demand and cached. Order-only dep on `bins` so an updated
+# binary doesn't force traffic/reputation to regenerate.
+$(NET)/peacetime_traffic.csv: | bins
+	@echo ">> generating $(TRAFFIC_DURATION) of peacetime traffic (one-time, cached)..."
+	$(FWD) --network-dir $(NET) --duration $(TRAFFIC_DURATION)
+
+$(NET)/reputation.csv: $(NET)/peacetime_traffic.csv | bins
+	@echo ">> bootstrapping reputation snapshot from $(TRAFFIC_DURATION) of traffic (boosted to fill the full window)..."
+	$(REP) --network-dir $(NET) --allow-boost
+
+# The three slot/liquidity attacks. ATTACK = CLI name, DIR = results/<DIR> (Debug form).
+.PHONY: slow-slot-jam fast-slot-jam fast-liq-jam
+slow-slot-jam: ; @$(MAKE) --no-print-directory run ATTACK=slow-slot-jam DIR=SlowSlotJam
+fast-slot-jam: ; @$(MAKE) --no-print-directory run ATTACK=fast-slot-jam DIR=FastSlotJam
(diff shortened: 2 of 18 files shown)


GeorgeTsagk:
The `-slot` variants are generally way more attractive for an attacker. When you jam with liquidity instead, the upfront cost is huuuge and the decaying window forces you to re-buy your entry to the protected slots, which significantly compensates the target over the attack window.

GeorgeTsagk:
### Sim update

Finished a pair of 3-month simulations, demonstrating the `slow-slot` / `fast-slot` variants, and how the attack becomes "profitable" at that point (ratio would get even more favorable for attacker over time)

  | stat | slow-slot | fast-slot |
  |---|---|---|
  | hold pattern | one hold, 91 days | 85s × ~92,500 cycles |
  | honest revenue denied (*) | **134.70M msat (98.9%)** | **134.68M msat (98.8%)** |
  | attacker total cost | **70.67M msat** | **70.67M msat** |
  | one-time entry (**) | 70.67M | 70.67M |
  | sustaining (refills) (***) | **0 (0 refills)** | **0 (0 refills)** |
  | **cost / denied** | **0.53** | **0.53** |
  | target reputation pairs | 11 → **0** / 105 | 11 → **0** / 105 |
  | profitable? (cost/denied < 1) | **yes** | **yes** |

(\*) - vs peacetime revenue equivalent
(\*\*) - cost to acquire initial reputation to initiate attack (what's needed to occupy all slots)
(\*\*\*) - cost to maintain one-time entry, needed due to decaying nature of window

GeorgeTsagk:
So I had to sanity check the above experiment, which lead to an interesting find:

Why did the simulation "refill" reputation only once? (at the start of the experiment)

I expected for the refills to be on-going, especially for a 3-month simulation (which happens to be the reputation window half-life).

**The catch**: as we're denying honest traffic, the `incoming_revenue` is going down over time, **significantly faster** compared to the reputation decay (shorter window, ~12x faster). 

This means that fast-failing HTLCs (<90s, zero penalty) can really saturate the defense if they break past the 2-week (incoming rev window) mark.

I calculated how much a 1% upfront fee would change the picture here, and considering that it would add to the incoming_revenue as the attack gets carried out, it only makes things worse by ~0.1X, with the attack paying 0.64x the denied revenue.

79912 rounds, 193 HTLCs per round := 15.42M total contribution to incoming revenue by upfront fees, but given the short decaying window of the incoming revenue window it will only settle at ~1.7M. This isn't that great, given that the near-zero HTLC fee was clamped to the 1msat upfront fee floor. More profitable channels targeted by this jam would see this contribution as way less significant.

Maybe slots should be priced independently of the current HTLC's fee?

GeorgeTsagk:
> Maybe slots should be priced independently of the current HTLC's fee?

The price of the slot depends on what usually occupies the slot.
e.g a channel that usually routes 1btc HTLCs should perceive each slot as more valuable than a channel that routes 1k sat HTLCs.

Could charge based on a value-weighted decaying average of past HTLC fees.

carlaKC:
> I removed the base fee for this simulation, as that alone makes the slot attacks very expensive. Reputation based defense should work well with 0 base fees, 

Our recommendation for zero fee nodes would be to set some "default" fee policy (1000 base/1ppm) and run the reputation algorithm _as if_ they had that fee policy set. This rewards peers that have good forwarding histories with us, though of course can't promise compensation under attack (little we can do for nodes that opt in to not being compensated for use of their resources).

> which is a network trend as well

I was under the impression that the network was trending away from zero fees? But regardless, the above recommendation should work reasonably.

> Why did the simulation "refill" reputation only once? (at the start of the experiment)

Not certain that I understand what you mean by "refill". The simulator will:
a) Do a once-off bootstrap of reputation values, so that we can start with a network with some history
b) During the attack, continue to send honest payments through the network

> as we're denying honest traffic, the incoming_revenue is going down over time, significantly faster compared to the reputation decay (shorter window, ~12x faster).

We aggregate our `incoming_revenue` for two weeks over the last 12x two week periods to smooth shocks caused under attack. Do you have a graph you could show me?

**Re findings**: I'm not surprised that removing the base fee makes a slot jamming pretty trivial, since the key pricing mechanism that we depend on for low value HTLCs is the base fee. I'd be interested to see what these results look like for a non-zero base fee (or with the suggested "pretend we have a fee" approach outlined above).

GeorgeTsagk:
> Not certain that I understand what you mean by "refill". The simulator will:
a) Do a once-off bootstrap of reputation values, so that we can start with a network with some history
b) During the attack, continue to send honest payments through the network

I'm referring to the attacker's inflated-fee payment at the start of the attack. The "refill" was something I added to the simulation which is a recurrent inflated-fee payment that refills reputation when we drop below the threshold throughout the attack. 

> Re findings: I'm not surprised that removing the base fee makes a slot jamming pretty trivial, since the key pricing mechanism that we depend on for low value HTLCs is the base fee. I'd be interested to see what these results look like for a non-zero base fee (or with the suggested "pretend we have a fee" approach outlined above).

From an older run (before removing the base fee) the HTLC opportunity cost was pretty much dominated by the 1sat base fee, making the initial reputation entry amount too big to maintain. The reputation decay alone was enough to compensate the target (i.e attacker has to refill in order to sustain attack -> refill is more than denied revenue).

> We aggregate our incoming_revenue for two weeks over the last 12x two week periods to smooth shocks caused under attack. Do you have a graph you could show me?

rep-vs-threshold

Also the data in csv format: 
[fast-slot-3mo_rep-vs-threshold.csv](https://github.com/user-attachments/files/28957990/fast-slot-3mo_rep-vs-threshold.csv)

> run the reputation algorithm as if they had that fee policy set

Ah good point :100:

carlaKC:
> I'm referring to the attacker's inflated-fee payment at the start of the attack. The "refill" was something I added to the simulation which is a recurrent inflated-fee payment that refills reputation when we drop below the threshold throughout the attack.

Gotcha - thanks for clarifying!

> making the initial reputation entry amount too big to maintain. 

Too big for who to maintain? So long as other peers are able to get reputation and use some protected resources, then a high threshold is fine?

> The reputation decay alone was enough to compensate the target (i.e attacker has to refill in order to sustain attack -> refill is more than denied revenue).

Requiring that the attacks has to compensate us to continue attacking sounds like the mitigation is working as designed? I'm not clear on what the problem is here.

> Graph

The shape of this graph looks a little suspicious to me. How long are you "bootstrapping" the network for? For our averages to work properly you'll need 6 months of history*.

---

\* OFC would like to acknowledge that of course we won't have 6 months of history IRL when we spin up new channels, but I think it's most sensible to think about already established channels for attack scenarios. We can always come up with smart bootstrapping defaults.

GeorgeTsagk:
> Too big for who to maintain? So long as other peers are able to get reputation and use some protected resources, then a high threshold is fine?

> Requiring that the attacks has to compensate us to continue attacking sounds like the mitigation is working as designed? I'm not clear on what the problem is here.

Too big for the attacker to maintain. I should be direct: the current model does work very well against slot attacks if we consider a base fee of 1sat ✔️. That is exactly the reason why I thought of removing it when experimenting with the slot attack variants, they work quite well when it's absent. Then again, we can consider the "virtual" base fee just for the htlc opportunity cost, so using a 1sat base fee on jam simulations is fair.

> The shape of this graph looks a little suspicious to me. How long are you "bootstrapping" the network for? For our averages to work properly you'll need 6 months of history*.

Hmm I'm curious, which part of the graph looks suspicious?

The attacker introduces themselves right before the attack. Their reputation gets built in a single step right before the jamming starts, they aren't a part of the 6-month reputation build up.

The reason the values match initially is because the attacker acquires exactly the amount of reputation they need to deploy all their HTLCs. Then we see the difference in the decaying windows (incoming revenue drops faster).

> * OFC would like to acknowledge that of course we won't have 6 months of history IRL when we spin up new channels, but I think it's most sensible to think about already established channels for attack scenarios. We can always come up with smart bootstrapping defaults.

👍 Yeah bootstrapping can use some tricks

PR #133, opened 2026-06-26
Update jamln to use sim-ln's virtual time for speedup

carlaKC:
This PR points to a (currently PR'd) version of sim-ln that grants us access to a virtual clock (sped up using a tokio paused runtime).

This updates simulations that take hours to take seconds. I have run a few of our existing attacks on both main and this branch, and the numbers are within 2% of each other. Given that we don't have perfect determinism, this seems reasonable enough to me.

### DIFF
--- a/ln-simln-jamming/src/bin/reputation_builder.rs
+++ b/ln-simln-jamming/src/bin/reputation_builder.rs
@@ -52,6 +52,11 @@ async fn main() -> Result<(), BoxError> {
         .unwrap();
 
     let cli = Cli::parse();
+    let clock = Arc::new(SimulationClock::new(1)?);
+    run(clock, cli).await
+}
+
+async fn run(clock: Arc<SimulationClock>, cli: Cli) -> Result<(), BoxError> {
     let forward_params: ForwardManagerParams = cli.reputation_params.into();
 
     let network = NetworkType::new(&cli.network, cli.attack_type, cli.attacker_bootstrap)?;
@@ -114,7 +119,6 @@ async fn main() -> Result<(), BoxError> {
         (bootstrap_records, 0)
     };
 
-    let clock = Arc::new(SimulationClock::new(1)?);
     let reputation_clock = Arc::clone(&clock);
     let mut reputation_interceptor: ReputationInterceptor<BatchForwardWriter, ForwardManager> =
         ReputationInterceptor::new_for_network(
--- a/ln-simln-jamming/src/bin/reputation_builder.rs
+++ b/ln-simln-jamming/src/bin/reputation_builder.rs
@@ -22,6 +22,7 @@ use ln_simln_jamming::{
 };
 use log::LevelFilter;
 use simln_lib::clock::SimulationClock;
+use simln_lib::runtime::block_on_virtual_time;
 use simple_logger::SimpleLogger;
 
 #[derive(Parser)]
@@ -43,8 +44,7 @@ struct Cli {
     pub attacker_bootstrap: Option<Duration>,
 }
 
-#[tokio::main]
-async fn main() -> Result<(), BoxError> {
+fn main() -> Result<(), BoxError> {
     SimpleLogger::new()
         .with_level(LevelFilter::Debug)
         .with_module_level("simln_lib::sim_node", LevelFilter::Debug)
@@ -52,8 +52,11 @@ async fn main() -> Result<(), BoxError> {
         .unwrap();
 
     let cli = Cli::parse();
-    let clock = Arc::new(SimulationClock::new(SystemTime::now()));
-    run(clock, cli).await
+
+    let start_time = SystemTime::now();
+    block_on_virtual_time(start_time, |clock| run(clock, cli))??;
+
+    Ok(())
 }
 
 async fn run(clock: Arc<SimulationClock>, cli: Cli) -> Result<(), BoxError> {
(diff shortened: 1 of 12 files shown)


elnosh (commented):
I'm having issues running the slow-jam attack - I have been running it for hours and still does not finish.

The behavior of the single-threaded paused tokio runtime is still somewhat obscure to me but I think I'm hitting the issue because of this https://github.com/carlaKC/sim-ln/blob/96c8f6dc3e8bf9d54e60f56f344f1e8e2a11fd00/simln-lib/src/clock.rs#L28-L29 and for slow-jam attack we hold the HTLC for 2 weeks while jamming and then we poll in a loop here https://github.com/carlaKC/jam-ln/blob/e7a9b6bda12017908f845c48fe257c178679cfa0/ln-simln-jamming/src/attacks/slow_jam.rs#L411-L420 until we are done holding the payment. Based on those comments in the sim-ln PR it seems this loop can be problematic.

carlaKC:
>  Based on those comments in the sim-ln PR it seems this loop can be problematic.

Discussed offline: loop creates futures that can be in `Pending` state, so we're ok here. Also pushed some small (unrelated) fixups to make sure we drop the lock before the sleep.

Speedup we get here is really nice, down from 20 minutes to a few seconds:
```
time cargo run --release --bin ln-simln-jamming -- \
    --network-dir networks/ln_slow_jam --attack-type slow-jam --attacker-bootstrap 10d \
    --reputation-margin-msat 0 --reputation-margin-expiry-blocks 10 \
    --target-reputation-percent 1 \
    > run.log 2>&1

cargo run --release --bin ln-simln-jamming -- --network-dir  --attack-type     95.54s user 14.62s system 40% cpu 4:33.86 total
```

One thing that we do lose here (which is a bit of a shame) is that we can't parallelize reading in the large peacetime file. I'm going to work on some changes to how we handle all this unweildy data anyway, so hopefully that's not such an issue long term.

carlaKC:
Diff since last review:
* [Change file parsing and fix simulation start time](https://github.com/carlaKC/jam-ln/compare/ffaf8b7f6c7fdf1113339e3311cec08f148961b3..1ed0029b1cf44d3beff7ac1eabfade5eadfac9a7)
* [Test fixup](https://github.com/carlaKC/jam-ln/compare/1ed0029b1cf44d3beff7ac1eabfade5eadfac9a7..adc931b344769eb6c78864cb64b7826a6ac9e73f)

ln-simln-jamming/src/parsing.rs
```diff
@@ -689,68 +689,31 @@ pub async fn history_from_file(
...
-            let incoming_add_ts: u64 = first_record[4].parse()?;
-            Some(incoming_add_ts.add(duration.as_nanos() as u64))
-        } else {
-            None
-        }
-    };
-
-    let mut tasks: Vec<tokio::task::JoinHandle<Result<Vec<BootstrapForward>, BoxError>>> =
-        Vec::with_capacity(num_chunks);
+    let mut csv_reader = csv::Reader::from_reader(BufReader::new(file));
```
elnosh: yeah I realized after first couple runs that this was slower because it's done single-threaded in the runtime sim-ln creates :( 

I don't think there's any reason we can't parallelize the read before actually running the simulation but that will require changes to sim-ln and for how much faster the simulation is already running I'd be fine if left as-is.

ln-simln-jamming/src/bin/reputation_builder.rs
```diff
@@ -43,17 +44,22 @@ struct Cli {
...
 
     let cli = Cli::parse();
-    let clock = Arc::new(SimulationClock::new(SystemTime::now()));
-    run(clock, cli).await
+
+    // Run on sim-ln's paused, single-threaded virtual-time runtime so the reputation replay completes instantly.
+    // Anchor virtual time at the real wall clock because the routing graph rejects channel updates whose timestamps
+    // are far from real time.
+    let start_time = SystemTime::now();
+    block_on_virtual_time(start_time, |clock| run(clock, cli))??;
```
elnosh: I'm wondering why does the reputation-builder need to be ran with the single-threaded runtime from sim-ln? Since we are not running a simulation but rather just replaying events. Maybe to keep the reputation building consistent but there's not much sleep/waits in those events

elnosh (commented):
cool! simulation ran in ~100s in release mode

carlaKC:
[Rebased + cleaned up a bunch of claude slop comments](https://github.com/carlaKC/jam-ln/compare/adc931b344769eb6c78864cb64b7826a6ac9e73f..e2809d13e384fbb68807825e19c2c20990629ec9)

+ [updated to sim-ln merged dep](https://github.com/carlaKC/jam-ln/compare/e2809d13e384fbb68807825e19c2c20990629ec9..9dab92bc8bebce1a33b8d8fae0e32491def83875)

ln-simln-jamming/src/parsing.rs
carlaKC: I have aspirations to blitz the `peacetime_traffic.csv` now that we can run sims faster (we can just run peacetime and attack networks in parallel) so shall address in a followup!

ln-simln-jamming/src/bin/reputation_builder.rs
carlaKC: Fair - probably unnecessary. I'm going to take a look at cleaning up all this setup so will get to this in a follow up!

elnosh (approved):

PR #134, opened 2026-07-07
Small cleanups to slowjam attack

carlaKC:
Investigating [hanging issue](https://github.com/carlaKC/jam-ln/pull/133#pullrequestreview-4612400628) from virtual clock, some low hanging fruit which I believe may be causing the sim to hang.


carlaKC:
Trivial cleanups, merged without review.

PR #136, opened 2026-07-14
ldk wrapper

elnosh:
I've ran a couple simulations with it. Will do some more testing before moving out of draft.

It is using changes from this [branch](https://github.com/elnosh/rust-lightning/commit/0f397e8fe90691c8c3a28aadcaa68775a8d3ede0) which mostly added added methods needed to impl the `ReputationManager`.

### DIFF
--- a/ln-simln-jamming/src/ldk_manager.rs
+++ b/ln-simln-jamming/src/ldk_manager.rs
@@ -1,0 +1,309 @@
+//! Wrapper that lets the LDK [`DefaultResourceManager`] be used as a drop-in
+//! [`ReputationManager`] for the simulator.
+
+use std::collections::HashMap;
+use std::sync::{Mutex, OnceLock};
+use std::time::Instant;
+
+use ln_resource_mgr::forward_manager::{ForwardManagerParams, SimulationDebugManager};
+use ln_resource_mgr::{
+    AccountableSignal, AllocationCheck, BucketResources, ChannelSnapshot, FailureReason,
+    ForwardResolution, ForwardingOutcome, HtlcRef, ProposedForward, ReputationCheck,
+    ReputationError, ReputationManager, ResourceCheck,
+};
+
+use lightning_ldk::ln::resource_manager::{
+    AllocationSnapshot, BucketSnapshot, ChannelReputationSnapshot, DefaultResourceManager,
+    ForwardingOutcome as LdkForwardingOutcome, ResourceManagerConfig,
+};
+use lightning_ldk::sign::EntropySource;
+
+/// The LDK manager tracks HTLC hold time as an absolute cltv expiry minus the height the HTLC was
+/// added at. The simulator's `ProposedForward` provides `expiry_in_height` as the number of blocks
+/// the HTLC may be held (the same value the native `ForwardManager` feeds into its risk
+/// calculation), so we pass it as the cltv expiry with a zero height added to reproduce the same
+/// hold-time and keep the two implementations comparable.
+const HEIGHT_ADDED: u32 = 0;
+
+const MAX_ACCEPTED_HTLCS: u16 = 483;
+
+/// A process-wide anchor mapping a reference [`Instant`] to a base unix timestamp. Shared by all
+/// adapters so reputation values remain comparable across nodes. The base is large enough that
+/// instants captured before the anchor (e.g. during bootstrap) still map to positive timestamps.
+static TIME_ANCHOR: OnceLock<(Instant, u64)> = OnceLock::new();
+
+const ANCHOR_BASE_UNIX_SECS: u64 = 4_000_000_000;
+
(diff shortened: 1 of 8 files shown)


ln-simln-jamming/src/ldk_manager.rs
```diff
@@ -0,0 +1,316 @@
...
+//!
+//! The two APIs differ in a few ways that this wrapper reconciles:
+//! - **Time**: the simulator works in [`Instant`]s; the LDK manager works in unix seconds. We map
+//!   between them using a single process-wide anchor so every node uses a consistent mapping.
+//! - **Channel sizing**: the simulator adds channels by `capacity_msat`; the LDK manager wants
+//!   `max_htlc_value_in_flight_msat` + `max_accepted_htlcs`. We map capacity to the in-flight
+//!   limit and use the protocol max of 483 accepted HTLCs.
+//! - **Snapshots**: the simulator seeds/reads reputation via `ChannelSnapshot`; we thread this
+//!   through the LDK manager's `add_channel_with_reputation` / `list_channels`. The LDK snapshot
+//!   doesn't carry capacity, so we track that here.
```
carlaKC: nit: move onto the struct itself, file level docs will get outdated pretty quick

ln-simln-jamming/src/ldk_manager.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+fn bucket_from_snapshot(snapshot: BucketSnapshot) -> BucketResources {
+    BucketResources {
+        slots_used: snapshot.slots_used,
+        slots_available: snapshot.slots_available,
+        liquidity_used_msat: snapshot.liquidity_used_msat,
+        liquidity_available_msat: snapshot.liquidity_available_msat,
+    }
+}
```
carlaKC, line 75: implement `From` or `Into`?

ln-simln-jamming/src/ldk_manager.rs
```diff
@@ -0,0 +1,316 @@
...
+                if snapshot.capacity_msat != capacity_msat {
+                    return Err(ReputationError::ErrChannelCapacityMismatch(
+                        capacity_msat,
+                        snapshot.capacity_msat,
+                    ));
+                }
+                self.inner.add_channel_with_reputation(
+                    channel_id,
+                    max_in_flight,
+                    MAX_ACCEPTED_HTLCS,
```
carlaKC, line 162: Maybe use the graph's `max_htlc_count` rather than a hard set default?

ln-simln-jamming/src/ldk_manager.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    fn remove_channel(&self, channel_id: u64) -> Result<(), ReputationError> {
+        self.capacities
+            .lock()
+            .map_err(|e| ReputationError::ErrUnrecoverable(e.to_string()))?
+            .remove(&channel_id);
+
+        self.inner
+            .remove_channel(channel_id)
+            .map_err(|_| ReputationError::ErrChannelNotFound(channel_id))
```
carlaKC: Tempted to leave this as `unimplemented` because we shouldn't use it in the sim.

carlaKC (commented):
Nice, just a few nits then this is ready to go.

Were the main changed needed for the LDK branch just utilities to get snapshots etc?

Also did you run any of the existing attacks with this, interested to know if there are any behavior changes.

elnosh:
> Were the main changed needed for the LDK branch just utilities to get snapshots etc?

yes, main reasons were:
- bucket snapshot that we return from `get_allocation_snapshot`
- list channels call that we use to get channel snapshots
- simulation helpers to mark general and congestion as jammed

ln-simln-jamming/src/ldk_manager.rs
elnosh: couldn't do it because `BucketSnapshot` and `ResourceManagerConfig` come from ldk

ln-simln-jamming/src/ldk_manager.rs
elnosh: should we fix separately? we use that hard default in `ForwardManager` too

elnosh:
> Also did you run any of the existing attacks with this, interested to know if there are any behavior changes.

I ran slow jam attack and saw similar results. Although that does not touch general and congestion functionality as those are hard set to "jammed"

carlaKC:
>  Although that does not touch general and congestion functionality as those are hard set to "jammed"

That's the case for most of the attacks, so I think it's fine! Thanks for re-running

ln-simln-jamming/src/ldk_manager.rs
carlaKC: yeah separately is good 👍

